PR #155, opened 2017-11-21
Cleanups related to JDK instructions

sposetti:
Also removed a line that mentions creating custom images for all platforms. That's not exactly true...so cleaned-that-up by removing.


lacikaaa:
Modifications conflicting with current master. They will be moved to another PR

PR #190, opened 2018-03-25
Ycloud debian9 and ubuntu16 docker images

leki75:


tibkiss:
+1, but I'd be more confident if Shane Kumpf could also take a look.

docker/debian9/Dockerfile
```diff
@@ -0,0 +1,82 @@
+FROM debian:9 as builder
```
skumpf: Note that multi-stage builds require a recent version of docker. Make sure where you plan to build supports this.

skumpf:
+1 lgtm

docker/debian9/Dockerfile
leki75: Thank you for your comment. At the moment we build docker images locally but contacted with @hortonworks/releng team about upgrading docker on jenkins slaves.

PR #192, opened 2018-04-02
BUG-98874 Extend prewarming with hdf stacks and mpacks

schfeca75:
@pdarvasi or @keyki pls review


saltstack/hortonworks/salt/pre-warm/tmp/install_hdp.sh
```diff
@@ -96,6 +122,9 @@ install_hdp() {
...
     fi
+    if [[ "$HDP_STACK_VERSION" == "3.1" ]]; then
```
keyki: We should check for STACK_TYPE too.

PR #234, opened 2018-07-26
SUST-1201 Fix Oracle JDK on Debian and Ubuntu

leki75:
* update apt cache at the start of the image build
* fix oracle jdk installation at non-interactive prompt
* install unlimited-jce-policy

### DIFF
--- a/saltstack/optional/pillar/oracle-java/init.sls
+++ b/saltstack/optional/pillar/oracle-java/init.sls
@@ -1,1 +1,6 @@
-JAVA_HOME: {{ salt['environ.get']('PREINSTALLED_JAVA_HOME') | default('/usr/java/default', True) }}
+JAVA_HOME:
+{% if grains['os_family'] == 'RedHat' %}
+  {{ salt['environ.get']('PREINSTALLED_JAVA_HOME') | default('/usr/java/default', True) }}
+{% else %}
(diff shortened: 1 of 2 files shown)


saltstack/optional/pillar/oracle-java/init.sls
```diff
@@ -1 +1,6 @@
-JAVA_HOME: {{ salt['environ.get']('PREINSTALLED_JAVA_HOME') | default('/usr/java/default', True) }}
+JAVA_HOME:
+{% if grains['os_family'] == 'RedHat' %}
+  {{ salt['environ.get']('PREINSTALLED_JAVA_HOME') | default('/usr/java/default', True) }}
+{% else %}
+  /usr/lib/jvm/java-8-oracle
```
keyki: Is this valid for SUSE too?

leki75: No, I have not tested it on SUSE.

lacikaaa: when we implemented oracle jdk support it was only for  Centos/Redhat 7. so when you select oracle optional state then it will fail anyway on Suse

keyki: Yes, I've talked with Gabor.

PR #257, opened 2018-11-29
Change rhel7 root device name in aws

daszabo:


pdarvasi:
I've made these changes in my last PR.

PR #277, opened 2019-02-04
BUG-112947 disable runnig dhclient-enter-hooks script on docker interface

lacikaaa:


keyki:
Can you please cherry-pick this commit to rc-2.9 as well. We're not going to re-build all images, but in case we need to the fix will be there.

PR #279, opened 2019-02-13
SUST-2327 Provide cloudbreak default/standard image for centos 7.6 fo…

arunmathew88:
…r ycloud usage

Change-Id: Ib66b684d2ab4da7e5da0ff5bc802d87fdf3f4c2b

### DIFF
--- a/docker/centos7.6/Dockerfile
+++ b/docker/centos7.6/Dockerfile
@@ -52,7 +52,7 @@ RUN rm -f /etc/salt/minion_id /etc/salt/pki/minion/minion.pem /etc/salt/pki/mini
 # restore systemctl to system default
 RUN ln -sf /usr/bin/systemctl.orig /usr/bin/systemctl
 
-ADD docker/centos7.3/files/image-runtime-scripts/start-services-script.sh /bootstrap/
+ADD docker/centos7.6/image-runtime-scripts/start-services-script.sh /bootstrap/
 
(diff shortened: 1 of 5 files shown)


docker/centos7.6/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    mv /usr/bin/systemctl /usr/bin/systemctl.orig && \
+    ln -sf /usr/bin/systemctl.py /usr/bin/systemctl
+
+RUN /tmp/saltstack/salt-setup.sh hortonworks
+RUN rm -f /etc/salt/minion_id /etc/salt/pki/minion/minion.pem /etc/salt/pki/minion/minion.pub
+
+# restore systemctl to system default
+RUN ln -sf /usr/bin/systemctl.orig /usr/bin/systemctl
+
+ADD docker/centos7.3/files/image-runtime-scripts/start-services-script.sh /bootstrap/
```
gnvk: Should this be centos7.6?

arunmathew88: Good catch, Thanks. It should be docker/centos7.6/image-runtime-scripts/start-services-script.sh

Actually centos7.4 -> 7.5 diff didn't have this change. But the file in centos7.3 seem to have the necessary changes for 7.5 as well. A bit misleading.

tibkiss:
LGTM

PR #284, opened 2019-04-01
DISTX-48. Burn fluentd install to CB CM prewarmed images.

oleewere:
- add fluent installation scripts
- will be included in the base image
- if OS is not supported (like Suse, i did not find installation script yet, probably we will need a custom way to install it) - if is not a fatal error (in order to not hurt any other base image creation)
- with setting `INCLUDE_FLUENT`, fluent installation can be skipped

Validated (but not on all ubuntu versions - let's assume the result is similar)

In Progress:
Move fluentd install to salt


oleewere:
@pdarvasi 
update notes for the new patch:
for some reason when im doing that in pillar (fluent/init.sls):
```bash
INCLUDE_FLUENT={{ salt['environ.get']('INCLUDE_FLUENT', 'Yes') }}
```
then used in the top.sls (of salt folder):
```bash
{% if pillar['INCLUDE_FLUENT'] == 'Yes' %}
    - fluent
{% endif %}
```
that is not worked ever, and i could not figure this out, so i have just used salt['environ.get'] without pillar part

PR #290, opened 2019-05-29
Fix debian9 image build

leki75:
* update debian9 image to 9.9 on OpenStack
* do not install git-all because it installs apache2 which
  blocks nginx installation


lacikaaa:
Could you add jira issue to the commit message?

PR #295, opened 2019-06-27
CLOUD-2066. Add CCM client to CB images.

jadair-cloudera:
With these changes, the CCM client infrastructure and prerequisites
have been added to the Salt configuration for CB images.

### DIFF
--- a/saltstack/base/salt/ccm-client/cdp/bin/reverse-tunnel.sh
+++ b/saltstack/base/salt/ccm-client/cdp/bin/reverse-tunnel.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -ux
+
+SCRIPT=`basename "$0"`
+ROLE=$1
+
+. /cdp/bin/reverse-tunnel-values-${ROLE}.sh
+
+# Setting gate time to 0 to retry on SSH failures due to connectivity issues on first try
+export AUTOSSH_GATETIME=0
+export AUTOSSH_LOGLEVEL=7
+export AUTOSSH_LOGFILE="/var/log/autossh-${ROLE}.log"
+# Set max backoff interval in case of recurring connection failures.
+export AUTOSSH_POLL=30
+
+PRIVATE_KEY=/tmp/pk.key
+
+if grep -q "BEGIN" ${ENCIPHERED_PRIVATE_KEY}; then
+    cat ${ENCIPHERED_PRIVATE_KEY} > ${PRIVATE_KEY}
+else
+    IV=436c6f7564657261436c6f7564657261
+    cat ${ENCIPHERED_PRIVATE_KEY} | openssl enc -aes-128-cbc -d -A -a \
+        -K $(xxd -pu <<< $(echo ${TUNNEL_INITIATOR_ID} | cut -c1-16) | cut -c1-32) \
+        -iv ${IV} > ${PRIVATE_KEY}
+fi
+
+chmod 400 ${PRIVATE_KEY}
+
+if [[ ${ROLE} == "SSH" ]]; then
+    LOCAL_IP=127.0.0.1
+    USER=${TUNNEL_INITIATOR_ID}_${ROLE}
+else
+    LOCAL_IP=0.0.0.0
+    USER=${TUNNEL_INITIATOR_ID}_${ROLE}
+fi
+
+exec autossh -M 0 -o "ServerAliveInterval 30" -o "ServerAliveCountMax 3" \
+-o UserKnownHostsFile=${PUBLIC_KEY} -N -T -R ${LOCAL_IP}:0:localhost:${HOST_PORT} \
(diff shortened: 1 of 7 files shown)


jadair-cloudera:
FYI, here is the client package module based on the code that is currently running in Altus. I renamed some of the scripts for consistency and got everything into Salt, but there are no functional changes.
https://github.infra.cloudera.com/thunderhead/thunderhead/tree/master/clients/autossh-adventure

saltstack/base/salt/ccm-client/cdp/bin/update-reverse-tunnel-values.sh
```diff
@@ -0,0 +1,16 @@
...
+CCM_SSH_PORT=${CCM_SSH_PORT}
+HOST_PORT=${HOST_PORT}
+ROLE=${ROLE}
+HOST=${HOST}
+TUNNEL_INITIATOR_ID=${TUNNEL_INITIATOR_ID}
+ENCIPHERED_PRIVATE_KEY=${ENCIPHERED_PRIVATE_KEY}
+CLOUD_PROVIDER=${CLOUD_PROVIDER}
+PUBLIC_KEY=${PUBLIC_KEY}
+EOF
+systemctl enable tunnel@{ROLE}.service
```
pdarvasi: This should be done via salt: https://docs.saltstack.com/en/latest/ref/states/all/salt.states.service.html#salt.states.service.running

saltstack/base/salt/ccm-client/init.sls
```diff
@@ -0,0 +1,19 @@
...
+  file.managed:
+    - name: /cdp/bin/reverse-tunnel-values.sh
+    - source: salt://{{ slspath }}/cdp/bin/reverse-tunnel-values.sh
+
+/cdp/bin/update-reverse-tunnel-values.sh:
+  file.managed:
+    - name: /cdp/bin/update-reverse-tunnel-values.sh
+    - source: salt://{{ slspath }}/cdp/bin/updatereverse-tunnel-values.sh
+
+/systemd/tunnel@.service:
```
pdarvasi: We should enable the service IMHO here: https://docs.saltstack.com/en/latest/ref/states/all/salt.states.service.html#salt.states.service.running

saltstack/base/salt/prerequisites/packages.sls
```diff
@@ -43,6 +43,9 @@ packages_install:
   {% endif %}
       - deltarpm
       - nvme-cli
+      - autossh
+      - openssl
+      - vim-common
```
pdarvasi: Is this one really needed?

saltstack/base/salt/ccm-client/cdp/bin/reverse-tunnel.sh
```diff
@@ -0,0 +1,38 @@
...
+export AUTOSSH_LOGLEVEL=7
+export AUTOSSH_LOGFILE="/var/log/autossh-${ROLE}.log"
+# Set max backoff interval in case of recurring connection failures.
+export AUTOSSH_POLL=30
+
+PRIVATE_KEY=/tmp/pk.key
+
+if grep -q "BEGIN" ${ENCIPHERED_PRIVATE_KEY}; then
+    cat ${ENCIPHERED_PRIVATE_KEY} > ${PRIVATE_KEY}
+else
```
pdarvasi: Do we have such a use-case when there is unencrypted ENCIPHERED_PRIVATE_KEY?

spacharya: Well technically in all production usages, we should be using Enciphered value, But in testing and debugging, depending on the team doing the work, they might want to use the plain key as is. That was the usecase. Also, Altus originally supported unencrypted and for backward incompatibility, we kept it as is.

saltstack/base/salt/prerequisites/packages.sls
spacharya: `$(xxd -pu <<< $(echo ${TUNNEL_INITIATOR_ID} |`

vim-commons provides the xxd funtionality which is used to decipher the enciphered text. If the xxd package is natively present, (i believe it is on ubunutu not on centos) then you can skip it. But for consistency, you might want to keep it around.

saltstack/base/salt/ccm-client/cdp/bin/reverse-tunnel.sh
jadair-cloudera: Peter, I will resolve this conversation. Reopen if you still have concerns.

saltstack/base/salt/prerequisites/packages.sls
jadair-cloudera: Peter, I will resolve this conversation. Reopen if you still have concerns.

pdarvasi: Ok, got it, thanks

PR #308, opened 2019-08-01
CB-2654. Add CCM setup to user-data-helper.sh

jadair-cloudera:
With these changes, the CCM client scripts have been cleaned
up and the CCM setup code has been added to user-data-helper.sh.


saltstack/base/salt/ccm-client/cdp/bin/reverse-tunnel.sh
```diff
@@ -14,25 +14,25 @@ export AUTOSSH_POLL=30
 
 PRIVATE_KEY=/tmp/pk.key
 
-if grep -q "BEGIN" ${ENCIPHERED_PRIVATE_KEY}; then
-    cat ${ENCIPHERED_PRIVATE_KEY} > ${PRIVATE_KEY}
+if grep -q "BEGIN" ${CCM_ENCIPHERED_PRIVATE_KEY_FILE}; then
+    cat ${CCM_ENCIPHERED_PRIVATE_KEY_FILE} > ${PRIVATE_KEY}
 else
     IV=436c6f7564657261436c6f7564657261
```
meowfaceman: Could you explain this bit of the code? From what I've read about IVs (initialization vectors), they shouldn't be reused. I know this wasn't your doing. We may want to look into replacing the fixed IV with something like:

```
IV=$(uuidgen | sed "s/-//g")
```

But that may be overkill if this is a test path.

jadair-cloudera: Because user data is potentially visible, we encrypt the private key so that it is not exposed. This is where we decrypt it. If the IV for encryption and decryption have to match, we presumably can't choose a random one here, but I will take a follow-on task to investigate whether that is the case.

PR #310, opened 2019-08-07
Support multiple parcels and node.js support in pre-warmed images. 

echekanskiy:


echekanskiy:
Scripts including some commented out data. Please let me know if that data need to be removed.

PR #314, opened 2019-08-12
CB-2654. Additional script improvements.

jadair-cloudera:
With these changes, the CCM variables are correctly passed through
to the update script, and the provider-specific instance ID is used
as the default CCM tunnel initiator ID.

### DIFF
--- a/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
+++ b/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
@@ -122,25 +122,34 @@ setup_ccm() {
   : ${CCM_HOST:? required}
   : ${CCM_SSH_PORT:? required}
   : ${CCM_PUBLIC_KEY:? required}
-  : ${CCM_TUNNEL_INITIATOR_ID:? required}
+  : ${CCM_TUNNEL_INITIATOR_ID:="$INSTANCE_ID"}
   : ${CCM_ENCIPHERED_PRIVATE_KEY:? required}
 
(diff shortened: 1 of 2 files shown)


saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -158,6 +167,14 @@ main() {
         start_nginx
       fi
     fi
+
+    INSTANCE_ID=
+    if [[ "$CLOUD_PLATFORM" == "AWS" ]]; then
+      INSTANCE_ID="`wget -q -O - http://169.254.169.254/latest/meta-data/instance-id`"
+    elif [[ "$CLOUD_PLATFORM" == "AZURE" ]]; then
+      INSTANCE_ID="`wget --header="Metadata: true" http://169.254.169.254/metadata/instance/compute/vmId?api-version=2017-08-01&format=text`"
```
bhavanki: Does the quoting work correctly here? It might be better to use `$()` instead of backticks.

Also, does this command require the options `-q -O -` ?

jadair-cloudera: Double quotes inside backticks work as desired. I haven't been able to test the Azure size, but agree that those options seem desirable.

pdarvasi: wget gets installed as part of the prerequisites, for all providers except YCloud, so that should not be an issue

PR #326, opened 2019-09-26
CB-3649 Launching a larger cluster does not work on AWS

gergopapi2:


saltstack/base/salt/dhcp/etc/dhcp/dhclient-enter-hooks
```diff
@@ -69,12 +70,15 @@ make_resolv_conf() {
...
                 log "Reverse DNS by ${nameserver}: ${REVERSE_DNS}"
-                echo "$REVERSE_DNS" | grep -qc "${search}"
+                echo "$REVERSE_DNS" | grep -A 1 "ANSWER SECTION" | grep -qc "${search}"
                 if [[ $? -eq 0 ]]; then
                     set_hostname_from_dns "$REVERSE_DNS" "$IP" "$search"
                     break
                 else
                     log "$nameserver DNS server didn't return valid result for $IP"
+                    # restarting network will force the dhcp client to run again and call make_resolv_conf
+                    /etc/init.d/network restart > $NETWORK_RESTART_LOGFILE 2>&1
```
keyki: If we don't restart the network daemon what happens?

gergopapi2: I tried it on RHEL6: unfortunately nothing. Automatic run occurred 25 minutes later ...

gergopapi2: exit 1 did not have an effect of restarting the script, either.

saltstack/base/salt/dhcp/etc/dhcp/dhclient-enter-hooks
```diff
@@ -69,12 +80,15 @@ make_resolv_conf() {
...
                 log "Reverse DNS by ${nameserver}: ${REVERSE_DNS}"
-                echo "$REVERSE_DNS" | grep -qc "${search}"
+                echo "$REVERSE_DNS" | grep -A 1 "ANSWER SECTION" | grep -qc "${search}"
                 if [[ $? -eq 0 ]]; then
                     set_hostname_from_dns "$REVERSE_DNS" "$IP" "$search"
                     break
                 else
                     log "$nameserver DNS server didn't return valid result for $IP"
+                    # restarting network will force the dhcp client to run again and call make_resolv_conf
+                    call_network_restart
```
akanto: Can we do a bit differently? I am afraid that it will be stuck in this loop forever if there is some DNS resolution problem, and since the network is continuously restarting, therefore I think ssh-ig into the box is not possible.

Would it be possible to investigate a different approach, a bit less invasive?

PR #328, opened 2019-10-01
fix {image_name}.json generation for base images (no pre-warm parcels)

benyoka:
Testing: I am not yet familiar with the image building process, so I only tested it by executing pp_images.json.

It should now create a valid json with the parcels missing.


keyki:
Can you send this PR to rc-2.14?

keyki:
This change will be merged from 2.14

PR #330, opened 2019-10-03
imagecatalog metadata prewarm parcel json array should be null if env…

attilapalfi92:
… var is empty


keyki:
Can you change branch to 2.14?

PR #339, opened 2019-10-11
generate torrent file for each parcel

keyki:


saltstack/hortonworks/salt/pre-warm/torrent.sls
```diff
@@ -0,0 +1,12 @@
...
+  pkg.installed:
+    - pkgs:
+      - mktorrent
+
+create_torrent:
+  cmd.script:
+    - name: salt://pre-warm/tmp/create_torrent.sh
+    - output_loglevel: DEBUG
+    - timeout: 9000
+    - unless: ls /tmp/create_torrent.status
```
lacikaaa: this file is never created

PR #348, opened 2019-10-31
fix redhat7 vim-common installation

zsedert:


rmkellogg:
Build was successful.  Many thanks.

PR #349, opened 2019-11-04
CB-4237 install ipa-client and openldap

keyki:


pdarvasi:
@zsedert This might brake Suse and Debian, as we might need to validate the package names there.

PR #352, opened 2019-11-06
Validate Java only if it was required

horadla23:


horadla23:
tested with 'CUSTOM_IMAGE_TYPE=base make build-aws-centos7', burning was success, validation skipped

horadla23:
Oracle Java installation is broken currently, need to fix in different PR

PR #353, opened 2019-11-07
CB-4366 Nginx should allow underscores in header names

risdenk:
Nginx by default prevents underscores in header names.
The HTTP spec allows underscores in headers. This change
configures Nginx to allow underscores in headers. It needs
to be configured at the top level since there are multiple
Nginx server config blocks.

http://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers

Some additional background on underscores in headers can
be found here:

https://stackoverflow.com/questions/22856136/why-http-servers-forbid-underscores-in-http-header-names

Signed-off-by: Kevin Risden <krisden@apache.org>


keyki:
This went into master so the first set of images will come with rc-2.16

PR #358, opened 2019-11-19
CB-4473 Freeipa has postgres errors in AWS system log

horadla23:


horadla23:
burnt a freeipa image with this change into our cloudera aws account and started an environment with 2.16.0-dev.155 successfully, postgres running on the freeipa instance as a service, no sing of error logs mentioned in the ticket, 
Screenshot 2019-11-19 at 16 34 22

PR #360, opened 2019-11-25
Prevent overwrite of /etc/motd

rmkellogg:
If the /etc/motd has content then do not overwrite with vendor motd. Necessary to allow for inclusion of customer required motd.


rmkellogg:
Invalid

PR #361, opened 2019-11-25
Prevent overwrite of /etc/motd 

rmkellogg:
If the /etc/motd has content then do not overwrite with vendor motd. Necessary to allow for inclusion of customer required motd.


rmkellogg:
Can you merge into rc-2.9 branch as well?

pdarvasi:
@rmkellogg In such cases, the best practice is that you open the PR against rc-2.9 branch and it should be merged into master once the PR is merged. 
The cherry-pick is not possible directly because it conflicts with the rc-2.9 line.

PR #370, opened 2020-01-08
CB-4983 manually creating symlink for ycloud postgres service

lnardai:
- added docker push goal to makefile for ycloud
- systemctl reenable was failing in docker environemnt so now we manually create a symlink to the original postgres-9.6 service file
- service user for docker repository request is pending
(only tested with ycloud image)
- added yarn-private directory creation to Dockerfile as it was create to late in the process

### DIFF
--- a/saltstack/base/salt/postgresql/init.sls
+++ b/saltstack/base/salt/postgresql/init.sls
@@ -108,9 +108,16 @@ systemd-link:
     - repl: "[Install]\nAlias=postgresql.service"
     - unless: cat /usr/lib/systemd/system/postgresql-9.6.service | grep postgresql.service
 
+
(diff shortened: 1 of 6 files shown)


saltstack/base/salt/postgresql/init.sls
```diff
@@ -108,9 +108,9 @@ systemd-link:
     - repl: "[Install]\nAlias=postgresql.service"
     - unless: cat /usr/lib/systemd/system/postgresql-9.6.service | grep postgresql.service
 
-reenable-postgres:
+create-postgres-service-link:
```
lacikaaa: this will be effective for all images not just ycloud. Are you sure about this?

lnardai: Yes I know, I haven't tested those path, It might be safer to create a separate branch for it.
I'll use the "yarn-private directory check" to decide.

PR #378, opened 2020-01-21
Govcloud fix - adjust AWS EC2 instance type

rmkellogg:
Adjust the AWS EC2 instance type used to supported allowed values in Govcloud environment.


rmkellogg:
The original t2.2xlarge is not supported in the us-gov-east-1 region.

rmkellogg:
Please merge to master and rc2.9.

pdarvasi:
@rmkellogg Normally we commit 1 squashed commit onto rc-2.9 and that is merged back to master. I will do it now, thanks for the commit.

PR #379, opened 2020-01-29
 CB-5301 Parallel image builds cause corrupted image catalog

horadla23:


horadla23:
@pdarvasi or @lacikaaa pls review, example https://github.com/hortonworks/cloudbreak-images-metadata/commit/72017dab587b1fa01c74ddf0611cc556e55b5904

Makefile
```diff
@@ -216,7 +218,8 @@ build-aws-ubuntu16:
 
 build-aws-centos7: export IMAGE_NAME := $(IMAGE_NAME)
 
-build-aws-centos7: build-aws-centos7-base
+build-aws-centos7: 
+	@ METADATA_FILENAME_POSTFIX=$(METADATA_FILENAME_POSTFIX) make build-aws-centos7-base
```
lacikaaa: shouldn't we add this to other OS types also? I know we don't burn those right now, but if we will at some point we might forget about this.

horadla23: it will also work for other make targets, the extra variable needed only here because we are executing another make target only here and we need to ensure that this postfix will be the same when `make build-aws-centos7-base` are running

pdarvasi:
@horadla23 I would have merged it in a minute....;)

horadla23:
@pdarvasi I did not know that :D thanks anyway!:)

PR #383, opened 2020-02-11
Upgrade pip to overcome zipp issue

horadla23:
failed: http://build.eng.hortonworks.com:8080/job/cloudbreak-multi-packer-image-builder-salt-v3/365/console
working build started with the feature branch of this PR: http://build.eng.hortonworks.com:8080/job/cloudbreak-multi-packer-image-builder-salt-v3/366/console

PR #394, opened 2020-03-27
wip: [build-gov-aws-amazonlinux] change path to pip

dekimsey:
In #383, the yum-managed pip is upgraded with pip. As part of that
upgrade pip removes /usr/bin/pip and install /usr/local/bin/pip. This
will likely break behavior of any root processes trying to use pip or
even yum updates of the pip package. That being said, we update our prep
script to resolve the new location so that it can get salt ready to go.

### DIFF
--- a/scripts/salt-install.sh
+++ b/scripts/salt-install.sh
@@ -6,9 +6,12 @@
 set -e -o pipefail -o errexit
 
 function install_salt_with_pip() {
-  pip install --upgrade pip
-  pip install virtualenv
+  # Upgrading pip replaces the system pip, removing /usr/bin/pip and installing
+  # /usr/local/bin/pip in its stead
+  pip install -v --upgrade pip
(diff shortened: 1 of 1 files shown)


pdarvasi:
@dekimsey Pls. open the PR agains rc-2.9 instead of master as CB 2.9.2 is the GA version which supports Ambari+HDP on Amazonlinux, thanks!

dekimsey:
Understood, we'll give the rebuilds a try and see what happens

dekimsey:
I'm going to close this MR. At least for us FIPS is a requirement and at this time Amazon Linux doesn't (at all) support FIPS. So there's zero reason to fix or attempt to use amazon linux in the usgov stacks.

I'll likely open a new PR with the changes I had to perform to get a RHEL 6 running with FIPS. The changes are a bit messier than I'd like so I'll be opening it to share the work with the expectation there will probably need to be some changes.

PR #410, opened 2020-05-22
CB-7191 | Fixing file permissions on reverse tunnel scripts to not ma…

dvguruprasad:
…ke them world readable


lacikaaa:
I'm going to fix it in a separate pr

PR #426, opened 2020-07-14
CB-7443. Changes to make some improvements to the time it takes to scale

sidseth:
up nodes.
- Introduces 2 pre-populates salt 'roles' in prewarmed images / images
containing fluent.
- Archives some components at the end of image creation, and sets up a
  script to uncompress these on first time image initialization. This is primarily
  to work around inefficiencies w.r.t retrieving multiple (thousands) of
  files during instance bootup vs a few archives.


sidseth:
Works along with hortonworks/cloudbreak-images#426, but both of these PRs can work independently.
Tested against CB-2.25-b124. Will tests against master after fixing review comments.

saltstack/base/salt/prerequisites/pip.sls
```diff
@@ -34,6 +34,12 @@ install_pyyaml:
     - name: pip install PyYAML --ignore-installed
     - unless: pip list | grep -E 'PyYAML'
 
+install_psycopg2:
```
keyki, line 37: Why is this needed? We installing psycopg2 with yum now and with version 2.8.5

scripts/salt-setup.sh
```diff
@@ -46,6 +46,53 @@ function apply_optional_states {
   fi
 }
 
+# This adds entities to the 'roles' key to be used during actual cluster install. Hence, this does not use the temporary config location.
+function add_single_role_for_cluster_salt {
+  local role=${1}
+  ${SALT_PATH}/bin/salt-call --local grains.append roles ${role} -l info --log-file=/tmp/salt-build-${saltenv}.log --log-file-level=debug
```
keyki, line 52: Grains are stored in the filesystem somewhere /etc/salt-master/ so it's easier just to create the file.

saltstack/base/salt/prerequisites/pip.sls
sidseth: This was added because of https://github.com/hortonworks/cloudbreak/blob/master/orchestrator-salt/src/main/resources/salt/salt/cloudera/agent/init.sls#L14

I don't see psycopg2 being installed anywhere in the image.

scripts/salt-setup.sh
sidseth: These get stored in /etc/salt/grains, which is where the grains set up dynamically by CB are also stored.

Would prefer to leave it via the salt invocation, so as not to worry about formats, existing contents of the file - and allow for future additions without worrying about formatting. Let salt take care of all of that.

saltstack/base/salt/prerequisites/pip.sls
lacikaaa: cm agent install brings it as a package dependency

saltstack/base/salt/cloud-init/etc/scripts/extract.sh
```diff
@@ -0,0 +1,58 @@
...
+# - etc.
+#
+# Downloading single large archives seems to be faster than individual files. (python-* - 2000+ files each, salt - 3000+, td-agent - 50K+ iirc)
+# Not all of these files are accessed while starting up, but a fair number are.
+# Using an archive in this manner, for a single node leads to the following in terms of startup cost.
+#
+# extracting the files -> +5 seconds
+# ipa install -> -5 to -13 seconds
+# salt startup -> -15 seconds
+# td-agent startup -> -5 seconds (td-agent - we likely need to purge unnecessary rubbish (man pages, unneeded plugins) in the image baking process)
```
lacikaaa, line 57: this means roughly 30sec gain totally
my opinion this isn't worth it. the possibility it breaks something vs the 30sec is just too big risk to take

scripts/salt-install.sh
```diff
@@ -89,7 +89,7 @@ function install_python_pip() {
     yum install -y python27-devel python27-pip
   elif [ "${OS_TYPE}" == "redhat7" ] || [ "${OS_TYPE}" == "amazonlinux2" ] ; then
     echo "Installing python36 with deps"
-    yum install -y python36 python36-pip python36-devel python36-setuptools
+    yum install -y python36 python36-pip python36-devel python36-setuptools vim strace
```
lacikaaa, line 92: why do we need vim and strace?

scripts/salt-setup.sh
```diff
@@ -46,6 +46,53 @@ function apply_optional_states {
...
+function create_archives_and_delete_files() {
+  tar -C /usr/lib64 -cvzf /usr/lib64/python3.6-archive.tar.gz python3.6
+  tar -C /usr/lib -cvzf /usr/lib/python3.6-archive.tar.gz python3.6
+  tar -C /usr/lib64 -cvzf /usr/lib64/python2.7-archive.tar.gz python2.7
+  tar -C /usr/lib -cvzf /usr/lib/python2.7-archive.tar.gz python2.7
+
+  # Salt (version as of this change: 3000.2) ends up taking a long time to load modules. vspere is an especially slow one taking 3 seconds.
+  # Salt does not seem to allow skipping module load ('disable_modules' only disables module usage, not module loading)
+  # So, deleting some modules which are not used, and tend to cause Exceptions / delays
+  find /opt/salt_3000.2 | grep modules | grep lxd | xargs rm
```
lacikaaa, line 79: these are a bit strange. why not use find for the whole?
`find /opt/salt_3000.2/lib/python3.6/site-packages/salt/modules/ -name "*lxd*"  -exec rm -f {} \;`

saltstack/base/salt/prerequisites/pip.sls
sidseth: Will remove it from here. Should I remove the section from agent/init.sls as well?

saltstack/base/salt/cloud-init/etc/scripts/extract.sh
sidseth: 30s is one of the bigger chunks that can be saved. Most other places are 1, 5, 10s - some of which get removed in the CB section of this patch. Could you please help me understand the risks here?
This change is limited to the image only - so is fairly easy to undo if required. It is configured to run via cloud-init once-per-instance, and I've verified that stop/start does not run it again. cloud-init has to complete before any of these components are used.
It does add some potential maintenance. e.g. if the salt or python version is changed, this may require some changes as well; my guess would be this isn't the only place.

scripts/salt-install.sh
sidseth: We don't. Makes the image a little more convenient to use. Can remove these.

scripts/salt-setup.sh
sidseth: No particular reason other than that's typically how I use find. Will change it over to to use find all the way.

saltstack/base/salt/prerequisites/pip.sls
lacikaaa: no, that's for backward compatibility

scripts/salt-setup.sh
```diff
@@ -46,6 +46,53 @@ function apply_optional_states {
...
+# This adds entities to the 'roles' key to be used during actual cluster install. Hence, this does not use the temporary config location.
+function add_single_role_for_cluster_salt {
+  local role=${1}
+  ${SALT_PATH}/bin/salt-call --local grains.append roles ${role} -l info --log-file=/tmp/salt-build-${saltenv}.log --log-file-level=debug
+}
+
+function add_prewarmed_roles {
+  if [ "${INCLUDE_FLUENT}" == "Yes" ]; then
+    # Note: This will need to be changed if making changes to versions etc in the prewarmed image.
+    local fluent_prewarmed="fluent_prewarmed_v1"
```
oleewere, line 58: can we het this value somehow differently from a config or something? it's a bit hidden in the code

saltstack/base/salt/cloud-init/etc/scripts/extract.sh
sidseth: Removed this for now. Will post a separate PR.

scripts/salt-setup.sh
sidseth: Moved it to the Makefile

scripts/salt-setup.sh
```diff
@@ -46,6 +46,37 @@ function apply_optional_states {
...
+function add_prewarmed_roles {
+  if [ "${INCLUDE_FLUENT}" == "Yes" ]; then
+    # Note: This will need to be changed if making changes to versions etc in the prewarmed image.
+    local fluent_prewarmed=${FLUENT_PREWARM_TAG}
+    echo "Adding ${fluent_prewarmed} to the list of roles for the final image"
+    add_single_role_for_cluster_salt ${fluent_prewarmed}
+  fi
+
+  if  [ "${STACK_TYPE}" == "CDH" -a ! -z "${CLUSTERMANAGER_VERSION}" -a ! -z "${CLUSTERMANAGER_BASEURL}" -a ! -z "${CLUSTERMANAGER_GPGKEY}" -a ! -z "${STACK_VERSION}" -a ! -z "${STACK_BASEURL}" -a ! -z "${STACK_REPOID}" ]; then
+    local prewarmed="prewarmed_v1"
```
lacikaaa, line 64: if the fluentd variable is extracted maybe this should be also

sidseth: Makes sense. Will make the change if the patches are going in.

PR #439, opened 2020-08-27
CB-8529: Starting Hue fails with longer stackname and 7.1.0 DE template

jszeles:
New httpd salt file is added to the prerequisites which removes the previously installed httpd and installs it again from a specified repository.


akanto:
Do we really trust repo.okay.com.mx ?

PR #438, opened 2020-08-25
CB-7443. Changes to make some improvements to the time it takes to scale

sidseth:
up nodes.
- Introduces 2 pre-populates salt 'roles' in prewarmed images / images
containing fluent.
- Remove some unused salt modules which are slow to load.

This makes 2 changes over the previous patch which was reverted. 1) Externalizes the prewarm role into the Makefile as requested by @lnardai. 2) Does not write the roles into the grains file directly. Instead writes these to a separate file. There's a corresponding PR on salt-bootstrap which takes care of populating this into the initial grains if required.

Sample output from a node using the custom salt-bootstrap, and this patch.
Compute node
```
ls /etc/salt/
grains  master.d  minion.d  minion_id  pki  prewarmed_roles
[root@sscdh2-compute3 salt]# cat /etc/salt/prewarmed_roles
fluent_prewarmed_v1
prewarmed_v1
[root@sscdh2-compute3 salt]# cat /etc/salt/grains
hostgroup: compute
recipes: []
roles:
- fluent_prewarmed_v1
- prewarmed_v1
- ipa_member
- manager_agent
- kerberized
- LOG_CLOUD_IDENTITY_ROLE
```

PR #441, opened 2020-09-08
CB-8123. Added metering prewarmed role support

oleewere:
applied changes from https://github.com/hortonworks/cloudbreak-images/pull/438 in order to add the role addition support


saltstack/base/salt/salt-bootstrap/init.sls
```diff
@@ -1,7 +1,7 @@
 install_saltbootstrap:
   archive.extracted:
     - name: /usr/sbin/
-    - source: https://github.com/hortonworks/salt-bootstrap/releases/download/v0.13.2/salt-bootstrap_0.13.2_Linux_x86_64.tgz
+    - source: https://github.com/hortonworks/salt-bootstrap/releases/download/v0.13.3/salt-bootstrap_0.13.3_Linux_x86_64.tgz
     - source_hash: md5=9cb97c667547ed98a6b7edcb4f61e987
```
keyki, line 5: the hash will not match

sidseth:
This ends up committing the changes in https://github.com/hortonworks/cloudbreak-images/pull/438, except for the deletion of the salt modules...
Curious about what testing was done? Base image is what was pending on the other PR (although I still think Azure is where the problem was). The module deletion would have had absolutely no impact, and instead the PR on salt-bootstrap is what would have fixed the issue seen after the previous commit which was reverted.

PR #450, opened 2020-10-02
CB-8502 adding google private subnet support

doktoric:


akanto (changes requested):
Looks good, but please also extend the main Readme.md with GCP instructions.

doktoric:
done

PR #453, opened 2020-10-13
Added cis-hardening checks

bormanbhaskar:
- Added checks to disable unwanted filesystem as per CB-8897.

### DIFF
--- a/saltstack/base/salt/cis-controls/init.sls
+++ b/saltstack/base/salt/cis-controls/init.sls
@@ -1,6 +1,7 @@
 #CentOS 1.1.1 Disable unused filesystems
 #https://jira.cloudera.com/browse/CB-8897
 
+{% if pillar['OS'] == 'centos7' %}
 {% set filesystems = ['cramfs', 'freevxfs', 'jffs2', 'hfs', 'hfsplus', 'squashfs', 'udf', 'fat'] %}
 
 {% for fs in filesystems %}
@@ -20,3 +21,4 @@
         - name: modprobe -r {{ fs }} && rmmod {{ fs }}
         - onlyif: "lsmod | grep {{ fs }}"
 {% endfor %}
+{% endif %}
--- a/saltstack/base/salt/cis-controls/init.sls
+++ b/saltstack/base/salt/cis-controls/init.sls
@@ -1,3 +1,6 @@
+#CentOS 1.1.1 Disable unused filesystems
+#https://jira.cloudera.com/browse/CB-8897
+
 {% set filesystems = ['cramfs', 'freevxfs', 'jffs2', 'hfs', 'hfsplus', 'squashfs', 'udf', 'fat'] %}
 
 {% for fs in filesystems %}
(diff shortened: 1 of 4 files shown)


saltstack/base/salt/top.sls
```diff
@@ -15,4 +15,5 @@ base:
     - fluent
 {% endif %}
     - ccm-client
+    - cis-controls
```
risdenk, line 18: @bormanbhaskar it might make sense to make this conditional based on an environment variable. I think this would make it easier to opt in/out during the build process. The fluent example is probably good to follow.

The rationale here is that it is much easier to change an environment variable - than it is to change the code if we need to revert this for any reason during testing.

risdenk (changes requested):

saltstack/base/salt/cis-controls/init.sls
```diff
@@ -0,0 +1,22 @@
+#CentOS 1.1.1 Disable unused filesystems
+#https://jira.cloudera.com/browse/CB-8897
```
keyki, line 2: Probably, it's worth to safeguard these changes to make them only on redhat7 based systems. We have lots of similar things in the code for debian,ubuntu,suse. Right now we don't support these, but it's a small investment that might worth it later.

saltstack/base/salt/custom/usr/local/bin/custom-ansible.sh
```diff
@@ -8,7 +8,7 @@ yum -y install ansible
 
 # Pull Ansible Playbooks from Git
 mkdir /tmp/ansible-playbooks
-git clone https://github.com/gitowner/ansible-playbooks.git /tmp/ansible-playbooks
+git clone https://github.com/cloudera/ansible-playbooks.git /tmp/ansible-playbooks
```
bormanbhaskar, line 11: #CB-9301: The existing example link was pointed to gitowner's account in which Cloudera doesn't have any control. Pointed it cloudera.

saltstack/base/salt/custom/usr/local/bin/custom-puppet.sh
```diff
@@ -14,7 +14,7 @@ yum install -y puppet-agent
 # Pull Puppet modules from Git
 # Note: It would be preferrable to use Puppetlabs r10k but baseline does not include required Ruby version
 mkdir /tmp/puppet-examples
-git clone https://github.com/gitowner/puppet-examples.git /tmp/puppet-examples
+git clone https://github.com/cloudera/puppet-examples.git /tmp/puppet-examples
```
bormanbhaskar, line 17: #CB-9301: The existing example link was pointed to gitowner's account in which Cloudera doesn't have any control. Pointed it cloudera.

saltstack/base/salt/top.sls
bormanbhaskar: Added conditional statement inside init.sls

saltstack/base/salt/cis-controls/init.sls
bormanbhaskar: Added target as centos7 for these changes.

risdenk:
PR #454 was created to split these changes out and not use `bormanbhaskar:master` - Sorry reviewers - I'll readd over there in a minute.

PR #454, opened 2020-10-14
CB-8897 - Disable unwanted filesystems

bormanbhaskar:
Added hardening checks as per https://jira.cloudera.com/browse/CB-8897

### DIFF
--- a/saltstack/base/salt/top.sls
+++ b/saltstack/base/salt/top.sls
@@ -15,5 +15,7 @@ base:
     - fluent
 {% endif %}
     - ccm-client
+{% if salt['environ.get']('INCLUDE_CIS') == 'Yes' %}
     - cis-controls
+{% endif %}
     - custom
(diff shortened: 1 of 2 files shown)


saltstack/base/salt/top.sls
```diff
@@ -15,4 +15,5 @@ base:
     - fluent
 {% endif %}
     - ccm-client
+    - cis-controls
```
risdenk, line 18:
```suggestion
{% if salt['environ.get']('INCLUDE_CIS') == 'Yes' %}
    - cis-controls
{% endif %}
```

This was lost on the move to the new PR. The idea here to be able to fully exclude the CIS benchmark stuff when building if necessary.

risdenk (changes requested):

saltstack/base/salt/top.sls
keyki: @bormanbhaskar can you please address Kevin's comment and I can merge the PR?

risdenk (approved):

PR #455, opened 2020-10-14
CB-9301 - The existing example link was pointed to gitowner's account

bormanbhaskar:


risdenk (changes requested):
Hmmm so I thought about this more after looking at it again - `gitowner` might be a variable that needs replacing by the end user? Its definitely not clear from the usage here if that is the case. It also doesn't help that `https://github.com/gitowner` is actually a valid user.

keyki (approved):

PR #468, opened 2020-11-10
CB-8482 Burn images on spot (AWS, GCP)

Bajzathd:
Enable spot on AWS (had to diable ENA, but it is not a problem because the source ami used has it already enabled).
Enable preemptible on GCP.


packer.json
```diff
@@ -61,6 +61,7 @@
     "salt_path": "{{ env `SALT_PATH` }}",
     "pyzmq_version": "{{ env `PYZMQ_VERSION` }}",
     "aws_instance_type": "t3.2xlarge",
+    "aws_spot_price": "0.3328",
```
akanto, line 64: Can we switch on spot without pricing?

Bajzathd: According to the documentation we should be able to, but when I tried it tried to start a normal instance and could not find its type - this was the only working spot scenario I could find.

akanto: Maybe we have an old version of packer?

Bajzathd: Tried it with the newest version and the same happened - so I rolled back the version change

akanto (approved):

PR #481, opened 2020-11-30
CB-10040. Remove port usage of 8888 from cdp-logging-agent.conf sample

oleewere:
details:
that is only used if salt does not override the sample configuration.
this can happen if not telemetry/logging is filled in the requests


biharitomi (approved):

biharitomi:
cc: @pkedvessy

oleewere:
converted to draft, maybe i can test out what if the rpm post install script does not contain any systemctl enable call

pdarvasi (approved):

pkedvessy (approved):

PR #482, opened 2020-11-30
CB-9982 Explicitly disable VM memory overcommit setting

cegganesh84:
1. In other Clouds this value is by default set to zero, which means overcommit based on heuristics.
2. In GCP we take a default Centos image from that cloud which already has a setting of 1, which means enable overcommit.
3. This commit brings GCP to the baseline of what we have in other clouds.


oleewere (approved):
lgtm

PR #490, opened 2020-12-12
CB-9183: [CCMv2] Make inverting proxy agent a part of ami/image

jszeles:
Salt stack is extended with install of ccmv2-inverting-proxy-agent.

### DIFF
--- a/saltstack/hortonworks/salt/ccmv2-inverting-proxy-agent/init.sls
+++ b/saltstack/hortonworks/salt/ccmv2-inverting-proxy-agent/init.sls
@@ -1,0 +1,6 @@
+/cdp/bin/ccmv2/inverting-proxy-agent:
+  file.managed:
+    - makedirs: True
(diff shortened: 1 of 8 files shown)


saltstack/hortonworks/salt/ccmv2-inverting-proxy-agent/init.sls
```diff
@@ -0,0 +1,6 @@
+install_ccmv2-inverting-proxy-agent:
+  cmd.script:
+    - name: salt://ccmv2-inverting-proxy-agent/install_ccmv2.sh
+    - output_loglevel: DEBUG
+    - timeout: 9000
+    - failhard: True
```
lacikaaa, line 6: file.managed source accepts urls. I think that would be a more elegant solution
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.managed
```
source

    The source file to download to the minion, this source file can be hosted on either the salt master server (salt://), the salt minion local file system (/), or on an HTTP or FTP server (http(s)://, ftp://).

lacikaaa (approved):

PR #496, opened 2020-12-22
CB-10271 - cloudera-scm user needs to be managed outside of CM

pdarvasi:
This PR implements 2 things:
1, creates static grain called builder_type based on env var PACKER_BUILDER_TYPE to be able to distinguish between cloud platform builds
2, populates ids jinja dictionary with the correct platform-specific uid and gid of cloudera-scm

Test:
Burnt non-default image with id 0edf9095-e87f-4f9e-8b32-066e2af1ca2e in dev catalog, validated with SDX and Data Hub launch.

### DIFF
--- a/saltstack/base/salt/prerequisites/user_uid.sls
+++ b/saltstack/base/salt/prerequisites/user_uid.sls
@@ -1,0 +1,34 @@
+{% set ids = salt['grains.filter_by']({
+    'amazon-ebs': {
+        'cloudera_scm_user': '992',
+        'cloudera_scm_group': '988',
+    },
+    'azure-arm': {
+        'cloudera_scm_user': '991',
+        'cloudera_scm_group': '987',
+    },
+    'googlecompute': {
+        'cloudera_scm_user': '992',
+        'cloudera_scm_group': '988',
+    },
+},
+grain='builder_type',
+default='amazon-ebs'
+)%}
+
+create_cloudera_scm_group:
+  group.present:
+    - name: cloudera-scm
+    - gid: {{ ids.cloudera_scm_group }} 
+
+create_cloudera_scm_user:
+  user.present:
+    - name: cloudera-scm
+    - fullname: Cloudera Manager
+    - shell: /sbin/nologin
+    - home: /var/lib/cloudera-scm-server
+    - createhome: False
+    - uid: {{ ids.cloudera_scm_user }}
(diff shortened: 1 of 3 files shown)


keyki (approved):

saltstack/base/salt/prerequisites/user_uid.sls
```diff
@@ -0,0 +1,34 @@
...
+    },
+},
+grain='builder_type',
+default='amazon-ebs'
+)%}
+
+create_cloudera_scm_group:
+  group.present:
+    - name: cloudera-scm
+    - gid: {{ ids.cloudera_scm_group }}
```
keyki, line 22: What happens on the YCloud image?

saltstack/base/salt/prerequisites/user_uid.sls
```diff
@@ -0,0 +1,34 @@
...
+  user.present:
+    - name: cloudera-scm
+    - fullname: Cloudera Manager
+    - shell: /sbin/nologin
+    - home: /var/lib/cloudera-scm-server
+    - createhome: False
+    - uid: {{ ids.cloudera_scm_user }}
+    - gid: {{ ids.cloudera_scm_group }}     
+    - groups:
+      - cloudera-scm
```
keyki, line 34: I would add a sanity check to make sure the user and group exist regardless of the builder_type. That way we can be sure that it works correctly even if the type changes or we add a new type.

pdarvasi: Makes sense, but the dictionary has a default clause meaning all new builders will inherit that setting

saltstack/base/salt/prerequisites/user_uid.sls
pdarvasi: It gets the default value, which is in this case, AWS. Do we have prewarmed YCloud images to check for the uid?

keyki: No, I just wanted to make sure the base image doesn't fail.

saltstack/base/salt/prerequisites/user_uid.sls
keyki: Right, I can merge without that check. Do you want me or should we wait for the Holidays to end?

keyki (approved):

PR #498, opened 2021-01-08
CB-10358: Changes to user-data-helper script to support one way TLS between inverting proxy agent and nginx

jszeles:
If CCMv2 is enabled while creating an environment, then ssl verify client parameter should be turned off and the nginx should listen in localhost:9443.

### DIFF
--- a/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
+++ b/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
@@ -116,6 +116,12 @@ setup_tls() {
     echo "cert-tool exists on the fs (backward compatibility)"
     create_certificates_cert_tool
   fi
+
+  if [[ "$IS_CCM_V2_ENABLED" == "true" ]]; then
+    echo "CCMv2 is enabled while creating an environment so ssl client verification is turned off and localhost:9443 used"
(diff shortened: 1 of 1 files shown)


lacikaaa (approved):

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -116,6 +116,12 @@ setup_tls() {
     echo "cert-tool exists on the fs (backward compatibility)"
     create_certificates_cert_tool
   fi
+
+  if [[ "$IS_CCM_V2_ENABLED" == "true" ]]; then
+    echo "CCMv2 is enabled while creating an environment so ssl client verification is turned off and localhost:9443 used"
+    sed -i "s/ssl_verify_client on;/ssl_verify_client off;/" /etc/nginx/sites-enabled/ssl-template
+    sed -i "s/listen       9443;/listen       localhost:9443;/" /etc/nginx/sites-enabled/ssl-template
```
lacikaaa, line 123: it might be better to use regexp for the spaces instead of the exact number of spaces. like if someone would delete a space in ssl-template it would break this and there is no test to validate this

lacikaaa: my comment was for both lines

lacikaaa (approved):

PR #507, opened 2021-01-28
CB-10902: Various fixes related to running the CCMv2 agent

TheTinkerDad:
- Fixed the proxy URL
 - Fixed the agent key deciphering command
 - Upgraded the agent to a newer version

### DIFF
--- a/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
+++ b/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
@@ -170,7 +170,9 @@ setup_ccmv2() {
 
   IV=436c6f7564657261436c6f7564657261
   AGENT_KEY_PATH=/etc/ccmv2/ccmv2-key.enc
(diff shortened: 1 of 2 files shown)


saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -184,8 +186,11 @@ setup_ccmv2() {
   chmod 400 "$TRUSTED_PROXY_CERT_PATH"
 
   INVERTING_PROXY_URL="$CCM_V2_INVERTING_PROXY_HOST"
+  
+  # A more sophisticated solution might need to be patched in later - tbh the script originally expected a full url with protocol scheme and closing slash
+  INVERTING_PROXY_FULL_URL="https://$INVERTING_PROXY_URL/"
```
TheTinkerDad, line 191: AFAIK this URL is coming from the java code - validating / extending the URL scheme should be done there, but for the time being and for this ticket, we can avoid the CB java code change.

PR #536, opened 2021-03-16
CB-10092: Adding an image for haproxy for YCloud loadbalancer support.

sxxgrc:
Jira: https://jira.cloudera.com/browse/CB-10092

This serves to add a new image to be used for YCloud datalake loadbalancers. I am curious as to how the image is put into the official Cloudera catalog, do I have to do anything for that or will it be done automatically?

Specifically, this will simply run the haproxy service, pointing at the gateway servers. I based this off the centos7 image and removed all of the extraneous Salt stuff. Please let me know if I can remove anything else to make this light.

This has been tested with a light duty  and a medium duty YCloud datalake and has worked. 

The test catalog that is in this branch will be deleted before I merge anything, I am just pointing my local cloudbreak at it for testing at the moment.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ build.sh
 !.dockerignore
 packer_no_pp.json
 test-folder
+*.code-workspace
--- a/Makefile
+++ b/Makefile
@@ -241,6 +241,10 @@ docker-build-centos76:
 	echo "Building image for ycloud2"
 	@ OS=centos7 OS_TYPE=redhat7 TAG=centos-76 DIR=centos7.6 make docker-build
 
+docker-build-yarn-loadbalancer:
+	echo "Building loadbalancer image for ycloud2"
+	@ OS=centos7 OS_TYPE=redhat7 TAG=yarn-loadbalancer DIR=yarn-loadbalancer make docker-build
+
 docker-build:
 	$(eval DOCKER_ENVS="OS=$(OS) OS_TYPE=$(OS_TYPE) SALT_VERSION=$(SALT_VERSION) SALT_PATH=$(SALT_PATH) PYZMQ_VERSION=$(PYZMQ_VERSION) PYTHON_APT_VERSION=$(PYTHON_APT_VERSION) TRACE=1")
 	$(eval DOCKER_BUILD_ARGS=$(shell echo ${DOCKER_ENVS} | xargs -n 1 echo "--build-arg " | xargs))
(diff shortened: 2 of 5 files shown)


sxxgrc:
One thing I recently noticed, is that the way the image selection logic works in CB is: it grabs all the images for the desired cloud provider (in this case Yarn) and then grabs the latest image out of those.
Naturally, if this new image is added as being for Yarn, that logic will select this new image for the datalake instances which is **not** desired.

As the loadbalancer logic we are implementing in CB will use the direct image ID/URL, I think it would be best to either: 
- not add this image to the catalog but only add it to the docker-sandbox.infra.cloudera.com repository
- or add this image to the catalog but not under the "yarn" cloud provider, maybe with "yarn-loadbalancer" as the provider

pkedvessy (commented):
I think we can add the image to the catalog without adding it to the default block. Image selection logic is working on the default images as far as I can remember.

sxxgrc:
> I think we can add the image to the catalog without adding it to the default block. Image selection logic is working on the default images.
> 
> What's the action item here? Merging this PR won't trigger anything. Ycloud images should be burnt manually...
> 
> If you don't want the image catalog being merged, why do we have it in the PR? :-) Please remove it.

Tbh I have no idea what the proper thing to do is for this, I just figured the PR would be the normal way to do it.
How do I burn it manually then, and add it to the catalog? Is that something you can do?
I'm happy to remove the PR, it's just that right now it is my only way of tracking this actually being done.

sxxgrc:
I went ahead and removed the catalog from the branch since I am done using it.

pkedvessy:
> How do I burn it manually then, and add it to the catalog? Is that something you can do?

Tbh I have never tried out ycloud image burning. So if you can burn it then please go ahead - as far as I know it is not more than call the new make target in the PR. Once we have the new image on dockerhub, image catalog needs to be updated manually in line with the previous ycloud images.

sxxgrc:
Thanks for the information! I'll try to figure out how to push it to the docker repo and let you know so we can merge this in for future use.
I am wondering about if adding the image to the catalog is even necessary though, since nothing will access the image through the catalog, but rather through its direct link through the docker repository. Do you have any thoughts on that @pkedvessy ?

pkedvessy:
> Thanks for the information! I'll try to figure out how to push it to the docker repo and let you know so we can merge this in for future use.
> I am wondering about if adding the image to the catalog is even necessary though, since nothing will access the image through the catalog, but rather through its direct link through the docker repository. Do you have any thoughts on that @pkedvessy ?

I can agree, If nothing needs to reference it from the catalog it is not necessary to update it.

sxxgrc:
@pkedvessy I finally was able to push in the docker image into the sandbox repo, so this can now be merged.

pkedvessy (approved):

PR #540, opened 2021-03-18
CB-10822:Tune default value of net.ipv4.tcp_retries2 to 7 ootb for cdp clusters

satyabollineni:
The value of net.ipv4.tcp_retries2 is changed to 7 which is 15 by default earlier. Screenshots for the value before and after the change (tested with base ami bake):

Before:
Screenshot 2021-03-18 at 8 12 45 PM

Post the change:

Screenshot 2021-03-18 at 8 12 57 PM

### DIFF
--- a/saltstack/base/salt/prerequisites/sysctl.sls
+++ b/saltstack/base/salt/prerequisites/sysctl.sls
@@ -49,3 +49,7 @@ vm.dirty_background_ratio:
 vm.dirty_ratio:
   sysctl.present:
     - value: 80
+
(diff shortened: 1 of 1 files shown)


cegganesh84 (approved):
Thank you for the testing.

TheTinkerDad (approved):
Approved, but won't merge yet - we have a lot of production hotfix builds ongoing and if this change causes any unforeseen problems, we'd have to delay all those hotfixes at once.

PR #543, opened 2021-03-23
CB-11695: Add FreeIPA consistency check to image

jszeles:
Checkipaconsistency tool is installed to freeipa images.


saltstack/freeipa/salt/ipaconsistency/init.sls
```diff
@@ -0,0 +1,3 @@
+install_checkipaconsistency:
+  cmd.run:
+    - name: pip install checkipaconsistency
```
lacikaaa, line 3: where would be the `checkipaconsistency` installed to? Into the salt virtualenv or outside of it? Have you managed to try it out?

lacikaaa (approved):

PR #546, opened 2021-04-09
CDPAM-1467 | Read CCMv2 agent params from a config file.

SachinHg:
In this PR I have made changes to create a .toml file and add the params required for running the ccmv2 agent on a node. This way we do not expose params when a non root user tries to read the command which might contain sensitive info like proxy username and passwords.
I have tested the changes manually on a freeIPA node and it works as expected.

I also started a build with the branch in http://build.eng.hortonworks.com:8080/job/cloudbreak-image-burn-salt-v3/16278/ which succeeded.


lacikaaa:
does the current version, which is on the images, support this config file?

SachinHg:
> does the current version, which is on the images, support this config file?

Thanks. I had misssed it :( . I have updated the image version and its SHA sum in the commit.

saltstack/base/salt/ccmv2-inverting-proxy-agent/cdp/bin/ccmv2/update-inverting-proxy-agent-values.sh
```diff
@@ -36,22 +36,23 @@ AGENT_CERT_PATH=$5
...
+scheme = "${SCHEME}"
+host = "${BACKEND_HOST}:${BACKEND_PORT}"
+trustedBackendCertificatePath = "${TRUSTED_BACKEND_CERT_PATH}"
+httpProxyURL = "${HTTP_PROXY_URL}"
 EOF
 
-if [ -f "$VALUES_FILE" ]; then
-    chmod 740 ${VALUES_FILE}
+if [ -f "$CONFIG_FILE" ]; then
+    chmod 740 ${CONFIG_FILE}
```
zabhi, line 55: Config file does not need to be executable. Can you check if 440 is enough for this?

saltstack/base/salt/ccmv2-inverting-proxy-agent/cdp/bin/ccmv2/update-inverting-proxy-agent-values.sh
```diff
@@ -36,22 +36,23 @@ AGENT_CERT_PATH=$5
...
-INVERTING_PROXY_URL=${INVERTING_PROXY_URL}
-BACKEND_ID=${BACKEND_ID}
-BACKEND_ADDRESS=${BACKEND_HOST}:${BACKEND_PORT}
-AGENT_KEY_PATH=${AGENT_KEY_PATH}
-AGENT_CERT_PATH=${AGENT_CERT_PATH}
-TRUSTED_BACKEND_CERT_PATH=${TRUSTED_BACKEND_CERT_PATH}
-TRUSTED_PROXY_CERT_PATH=${TRUSTED_PROXY_CERT_PATH}
-SCHEME=${SCHEME}
-HTTP_PROXY_URL=${HTTP_PROXY_URL}
+CONFIG_FILE=/cdp/bin/ccmv2/config-${BACKEND_ID}.toml
```
zabhi, line 39: Now that we know that we would only have a single agent running on a system, can we not suffix with backend id? `config.toml` should be enough for this.

saltstack/base/salt/ccmv2-inverting-proxy-agent/cdp/bin/ccmv2/run-inverting-proxy-agent.sh
```diff
@@ -6,7 +6,6 @@
 SCRIPT=`basename "$0"`
 BACKEND_ID=$1
 LOG_PATH="/var/log/ccmv2-inverting-proxy-agent.log"
+CONFIG_FILE="/cdp/bin/ccmv2/config-${BACKEND_ID}.toml"
```
zabhi, line 9: Now that we know that we would only have a single agent running on a system, can we not suffix with backend id? `config.toml` should be enough for this.

saltstack/base/salt/ccmv2-inverting-proxy-agent/cdp/bin/ccmv2/update-inverting-proxy-agent-values.sh
zabhi: I would also recommend updating the service name to remove the backend id. (Not necessarily in the same PR) This would make debugging easier.

SachinHg: Ok. I thought about this but decided against in order to keep things in sync with how it is right now. Will make the changes.

saltstack/base/salt/ccmv2-inverting-proxy-agent/cdp/bin/ccmv2/update-inverting-proxy-agent-values.sh
SachinHg: You are right. Earlier it needed to be executable as we were running the values file. Now that is not required. So 640 should be good enough

saltstack/base/salt/ccmv2-inverting-proxy-agent/cdp/bin/ccmv2/run-inverting-proxy-agent.sh
```diff
@@ -6,7 +6,6 @@
 SCRIPT=`basename "$0"`
 BACKEND_ID=$1
 LOG_PATH="/var/log/ccmv2-inverting-proxy-agent.log"
+CONFIG_FILE="/cdp/bin/ccmv2/config.toml"
 
-. /cdp/bin/ccmv2/inverting-proxy-agent-values-${BACKEND_ID}.sh
-
-exec /cdp/bin/ccmv2/inverting-proxy-agent -scheme ${SCHEME} -proxy ${INVERTING_PROXY_URL} -host ${BACKEND_ADDRESS} -backend ${BACKEND_ID} -disable-gce-vm-header=true -trusted-proxy-cert-path ${TRUSTED_PROXY_CERT_PATH} -trusted-backend-cert-path ${TRUSTED_BACKEND_CERT_PATH} -cert-path ${AGENT_CERT_PATH} -key-path ${AGENT_KEY_PATH} -http-proxy-url=${HTTP_PROXY_URL} > "$LOG_PATH" 2>&1
+exec /cdp/bin/ccmv2/inverting-proxy-agent -config-file ${CONFIG_FILE} > "$LOG_PATH" 2>&1
```
lacikaaa, line 11: I think this should be like `>> "$LOG_PATH"` to ensure the log is not overwritten

SachinHg: Ok. I followed what was there already but the comment makes sense to me. Appending would be a better way of doing it since we do not want to lose the logs if the agent got restarted for some reason. I will make this change

lacikaaa (approved):

SachinHg:
Triggered http://build.eng.hortonworks.com:8080/job/cloudbreak-image-burn-salt-v3/16285/ with the changes.

PR #551, opened 2021-04-20
CDPAM-1531 | Add bash script for ccm debugging information collection

meetrajsinhvala:
This PR contains a bash script that will be used to collect the information required for ccm debugging and it will also perform record the output of few connectivity checks.

This will be placed along with autossh process launcher script and it is part of the ccmv1 debugging improvement tasks. Please provide your review.

### DIFF
--- a/saltstack/base/salt/ccm-client/cdp/bin/ccm_debugger.sh
+++ b/saltstack/base/salt/ccm-client/cdp/bin/ccm_debugger.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+
+BASE_DIRECTORY=$(cd $(dirname $0); pwd -L)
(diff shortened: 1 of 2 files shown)


lacikaaa (changes requested):
we already have cdp-doctor for similar reasons. please talk with @oleewere and integrate this into that one instead adding another script

oleewere:
@pdarvasi , just note for you: we have discussed this already on slack, the steps will be implemented in cdp-doctor (most of all these checks have been implemented already in cdp-doctor, but not all of these...like collecting specific logs by the cdp-doctor command itself etc.)

lacikaaa:
closing this based on the comment from @oleewere

PR #552, opened 2021-04-22
CB-12248 Set the net.ipv4.tcp_keepalive_time to 300 to avoid rare NLB…

cegganesh84:
… timeouts

1. In DE HA cluster there is an occasional timeout for beeline queries.
2. NLB idle-timeout is 350s compared to the system keepalive time of 2 hours in manager node.
3. To make sure the clients never get an RST because the NLB connection was idle-timeout reduce keep alive timeout value below 350.

No testing done, based on an analysis from a blog post https://medium.com/tenable-techblog/lessons-from-aws-nlb-timeouts-5028a8f65dda

### DIFF
--- a/saltstack/base/salt/prerequisites/sysctl.sls
+++ b/saltstack/base/salt/prerequisites/sysctl.sls
@@ -14,6 +14,10 @@ net.ipv4.ip_local_port_range:
   sysctl.present:
     - value: "32768 61000"
 
+net.ipv4.tcp_keepalive_time:
+  sysctl.present:
+    - value: 300
+
 net.core.netdev_max_backlog:
    sysctl.present:
     - value: 20000


hreeve-cloudera:
@cegganesh84 A change from 7200s to 300s seems fairly significant. Are here any concerns about potential side effects of making this change?

cegganesh84:
@hreeve-cloudera Yes, this will affect all FreeIPA, datalake, and datahub clusters. Unfortunately, it is hard to know where it will affect. It is hard to test it unless we let it go through the QE cycle for 7.2.10. I have never seen a value of keep alive being more than 5 mins. But a datamart cluster being accessed via BI tool that expects a long-running query of more than 1-hour being affected is also not tested by anyone at the moment. 

An alternative to this for reducing the blast radius is to set this value only for DE HA cluster knox instances. @moresandeep do you know if Knox can take a keep alive setting for the application instead of relying on the OS?

moresandeep:
@cegganesh84 Yes, knox does have settings to adjust socket and connection timeouts. These are the properties I think you are looking for

- `httpclient.socketTimeout ` - The amount of time to wait for data on a socket before aborting the connection. The natural unit is milliseconds, but a ‘s’ or ‘m’ suffix may be used for seconds or minutes respectively. The default timeout is system dependent but is likely to be indefinite. Default 20s

-  `httpclient.connectionTimeout ` - The amount of time to wait when attempting a connection. The natural unit is milliseconds, but a ‘s’ or ‘m’ suffix may be used for seconds or minutes respectively. The default timeout is system dependent.  Default 20s

This is an example of how to set it at service level
```
<service>
     <role>HIVE</role>
     <param>
         <name>httpclient.socketTimeout</name>
         <value>180s</value>
     </param>
</service>
```
Ref: https://knox.apache.org/books/knox-1-5-0/user-guide.html#Common+Service+Config

Having said that, there are other timeouts to consider, i.e. the NLB timeout, the backend service timeout.

cegganesh84:
@moresandeep Thanks. Great point about other timeouts. NLB actually proxies knox request through nginx which already has a keep-alive timeout of 75s. I need to dig this a little deeper, because I can not explain the behaviour with that setting.

PR #553, opened 2021-04-23
CB-12256 Don't set /var/log/ directory permissions for CIS benchmark

risdenk:
The CIS v3.0.0 CentOS 7 benchmark recommends to set
directory permissions on subfolders in /var/log/. This
breaks system components that use folders like
/var/log/pki/ as in the case of ipa-server.

This change uses the command from the CIS v2.2.0
CentOS 7 benchmark to only set permissions on the files
under /var/log/.


risdenk:
@pkedvessy this has NOT been tested yet.

@bormanbhaskar this would be good to test early next week.

pkedvessy:
@risdenk Thanks a lot, I can take to ownership on testing the change since we dont have an e2e test job for it.

@bshashikant FYI ^

pkedvessy (approved):

PR #573, opened 2021-05-24
CDPAM-1721: Start using CCM v2 from RPMs in CB VM images

zabhi:
Switches mode to start consuming an RPM instead of raw binary.
Also starts using systemd unit service configuration from the RPM.
Updates configuration to latest format.

Other related changes:
- CDPAM-1647: Update setup scripts to set and pass relevant proxy parameters
    Starts _trying_ to read proxy configuration from EnvironmentFile at `/etc/cdp/proxy.env`.
    Does not fail if the values are absent.
- CDPAM-1658: Update cloudbreak-images to consume latest jumpgate-agent
    Updates agent version to latest

Build:
http://build.eng.hortonworks.com:8080/job/cloudbreak-multi-packer-image-builder-salt-v3/3441/consoleFull


saltstack/base/salt/ccm-jumpgate-agent/cdp/bin/ccmv2/generate-config.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -ux
+#
+# Copyright (C) 2020 Cloudera, Inc.
```
SachinHg, line 3: 2021 maybe?

lacikaaa:
I think the naming should be sync with the decision here: https://cloudera.slack.com/archives/C01D48PH4TE/p1621597838016100
so if the backend side will use ccmv2_jumpgate, then the folders in the images repo should reflect this also. or we should just stick with ccmv2 as all this will be installed for the clusters too, as it's in the `base` part.

I will convert this to draft just to make sure no one merges it accidentally

SachinHg (approved):

lacikaaa (approved):
please don't forget to squash your commits

PR #575, opened 2021-05-31
CB-12648 proxy.env permission fix

bergerdenes:


zabhi (approved):

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -206,6 +206,7 @@ setup_ccmv2() {
     if  [[ ! -z ${PROXY_NO_PROXY_HOSTS} ]]; then
       echo no_proxy=${PROXY_NO_PROXY_HOSTS} >> $PROXY_ENV_FILE
     fi
+    chmod 644 $PROXY_ENV_FILE
```
zabhi, line 209:
```suggestion
    chmod 640 $PROXY_ENV_FILE
```
This might contain proxy credentials.

zabhi (changes requested):
640 might be better here.

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
bergerdenes: but, this file is now created as `root` and the new jumpgate agent user will not be able to read this file -> no_proxy option is never applied

zabhi (approved):

TheTinkerDad (approved):

PR #590, opened 2021-06-24
CB-13209 Add custom nginx response header

bergerdenes:
Add a response header in our nginx so we can check if the request reached nginx,
or failed between CB - CCM


lacikaaa (changes requested):
same as for the other pr, please add `always`

lacikaaa (approved):

pkedvessy (approved):

PR #599, opened 2021-07-19
CB-13448 Workaround to sequentially mount disks with systemd init system

daszabo:


lacikaaa:
shall we add this only for AWS, shouldn't we?

saltstack/base/salt/mount/init.sls
```diff
@@ -0,0 +1,19 @@
+{% if grains['os'] == 'Amazon' %}
```
lacikaaa, line 1: we don't use amazon linux anymore, are you sure this will be amazon in case of centos?

daszabo: I am not, and I do not see any better way to check if the cloud platform is AWS or not. I should check whether the issue affects Azure VMs too, if it does, I could kill two birds with the same stone

daszabo:
The atlas_artifact_type variable's value, which I use to check if this is going to be an AWS image build, is set in the Makefile goal: build-aws-centos7-base to 'amazon', when we are building an aws image.

pdarvasi:
@daszabo Have you tried this with non-AWS burning? I have a concern about the state being empty causing a Salt error ..

pdarvasi:
@daszabo Btw, there is a grain for Provider distinction, pls see this:

https://github.com/hortonworks/cloudbreak-images/blob/584c59c15c2db84bf9820959d8da3d10a9c01f66/saltstack/base/salt/prerequisites/user_uid.sls#L2-L16

lacikaaa:
> @daszabo Have you tried this with non-AWS burning? I have a concern about the state being empty causing a Salt error ..

I think the if could be moved to the top file in that case

lacikaaa:
@daszabo have you been able to test this? is it ready for merge?

pdarvasi:
> > @daszabo Have you tried this with non-AWS burning? I have a concern about the state being empty causing a Salt error ..
> 
> I think the if could be moved to the top file in that case

I would rather have uses this than making top file more unreadable, but it is only personal preference: 
https://docs.saltproject.io/en/latest/ref/states/all/salt.states.test.html#salt.states.test.nop

pdarvasi:
I have tested for Azure successfully:

```    arm-centos7: [INFO    ] Fetching file from saltenv 'base', ** done ** 'mount/init.sls'
    arm-centos7: [INFO    ] Running state [NOP - Mount workaround only needed on AWS] at time 13:02:42.475911
    arm-centos7: [INFO    ] Executing state test.nop for [NOP - Mount workaround only needed on AWS]
    arm-centos7: [INFO    ] Success!
    arm-centos7: [INFO    ] Completed state [NOP - Mount workaround only needed on AWS] at time 13:02:42.476916 (duration_in_ms=1.009)
    arm-centos7: local:
    arm-centos7: ----------
    arm-centos7:           ID: nop-for-mount-workaround-on-other-platforms
    arm-centos7:     Function: test.nop
    arm-centos7:         Name: NOP - Mount workaround only needed on AWS
    arm-centos7:       Result: True
    arm-centos7:      Comment: Success!
    arm-centos7:      Started: 13:02:42.475907
    arm-centos7:     Duration: 1.009 ms
    arm-centos7:      Changes:
    arm-centos7:
    arm-centos7: Summary for local
    arm-centos7: ------------
    arm-centos7: Succeeded: 1
    arm-centos7: Failed:    0
    arm-centos7: ------------

lacikaaa (approved):

pdarvasi (approved):

PR #607, opened 2021-07-27
CB-13483 Start CCMv2 Jumpgate agent on FreeIPA host only

bergerdenes:
Refactor proxy config as CCMV2 setup will not run on other hosts.

### DIFF
--- a/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
+++ b/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
@@ -16,6 +16,12 @@ set
 : ${XARGS_PARALLEL:=}
 # : ${XARGS_PARALLEL:="-P 20"}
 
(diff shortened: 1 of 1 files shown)


lacikaaa (approved):
LGTM
one suggestion: the new logic wouldn't require the CCMv2 flag to be true in case you refactor this a bit:
https://github.com/hortonworks/cloudbreak-images/blob/master/saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh#L120-L129

something like
```
if [[ JUMPGATE ]]; then
  turn off mtls
elif [[ CCMv2 ]]; then
  turn off mtls
  bind to localhost
fi
```
this way the logic would be a bit more easy to read. and we won't need the CCMv2 flag to be set in case of jumpgate. but sure for backward compatibility we would have to keep that way for a while.

PR #625, opened 2021-08-31
CB-13906. Add sysstat (+ goaccess) packages

oleewere:


cegganesh84:
Oliver, +1 for sysstat package. But not sure about goaccess, never seen it before, and what kind of security compliance it has.

oleewere:
@cegganesh84 , you can generate reports like this from nginx access logs:
https://rt.goaccess.io/?20210826211303
of course I would only use that statically (so only during diagnostics) - this can be optional as the salt logs (where you need to apply '--include-salt-logs' if you want to collect that, same can be done here with '--include-nginx-report')

cegganesh84:
Thanks for the explanation, +1 if it will **not** be exposed in a port in the customer cluster.

oleewere:
ok, will open PR from draft once i have tested the image build

pdarvasi (approved):

PR #626, opened 2021-09-08
RHEL fixes

Bajzathd:
CDPR-44 Handle existing exclude directive in /etc/yum.conf
CDPR-43 Add sbin directories to path if they are not present
CDPR-42 Enable rhscl repo if not already enabled
CDPR-40 CDPR-41 Skip gpgchecks


TheTinkerDad (approved):
LGTM

PR #632, opened 2021-09-15
CDPR-43 CDPR-44

Bajzathd:
CDPR-43 Fix PATH variable
CDPR-44 Fix /etc/yum.conf append


eugene-cd (commented):
This fix does not seem to change the paths for the other sls files

PR #643, opened 2021-09-29
CDPR-52: Prevent dnsmasq from starting on RHEL VMs

jszeles:
Dnsmasq needs to be stopped and disabled otherwise unbound service will not start.


TheTinkerDad (approved):
LGTM - PS already tested this in the customer env.

PR #654, opened 2021-11-25
CB-14420 CB-14833 Enable Oozie HA and Knox HA failover testing in YCloud

cegganesh84:
1. For Oozie HA to work YCloud needs to listen on multiple ports.
2. For Knox HA failover testing to work the check port of the HAproxy should be Knox port 8443


hreeve-cloudera (approved):
LGTM

PR #663, opened 2022-02-03
CB-15450. Patch stageuser and user commands for better performance

handavid:
This commit patches user and stageuser commands to skip some
behaviors that take a long time but are not needed for our
use case.

stageuser_activate has been modified to:
  - not add user to default group (ipausers)
  - not run a query to populate the return value

user_disable and user_del have been modified to:
  - not check whether the user being disabled/deleted is the
    last enabled member of a protected group (i.e., admins).
    This check was performed regardless of whether the user
    is a member of the admins group or not. Usersync has
    protections to not modify the admin user, so this check
    is extraneous. A flag was added to this check so that the
    default behavior is maintained for callers outside of
    usersync.

user_add has been modified to:
  - not add user to the default group (ipausers)

### DIFF
--- a/saltstack/freeipa/salt/freeipa/init.sls
+++ b/saltstack/freeipa/salt/freeipa/init.sls
@@ -39,3 +39,17 @@ install_freeipa_healthagent_rpm:
 net.ipv6.conf.lo.disable_ipv6:
   sysctl.present:
     - value: 0
+
+/usr/lib/python2.7/site-packages/ipaserver/plugins/stageuser.py:
+  file.patch:
+    - source: salt://{{ slspath }}/tmp/stageuser.py.patch
+    - hash: md5=c34ee2a14a0480f07faef36507626bc6
+    - require:
+      - freeipa-install
+
+/usr/lib/python2.7/site-packages/ipaserver/plugins/user.py:
+  file.patch:
+    - source: salt://{{ slspath }}/tmp/user.py.patch
+    - hash: md5=47508b761dfe42f173eee53a90bfb4db
+    - require:
+      - freeipa-install
--- a/saltstack/freeipa/salt/freeipa/tmp/stageuser.py.patch
+++ b/saltstack/freeipa/salt/freeipa/tmp/stageuser.py.patch
@@ -1,0 +1,34 @@
+--- stageuser.py.orig	2022-01-31 13:33:25.989475331 -0600
++++ stageuser.py	2022-02-01 17:12:51.007022077 -0600
+@@ -727,26 +727,12 @@
+                              "active %s", active_dn)
+             raise
+ 
+-        # add the user we just created into the default primary group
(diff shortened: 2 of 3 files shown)


handavid:
@lacikaaa What's the normal procedure for testing these changes? I built myself an aws ami but had to temporarily hack around some configurations (region, vpc, etc) because I don't have access to the accounts y'all use to build images. I verified that salt did apply the patches properly.

lacikaaa:
@handavid please check: https://cloudera.atlassian.net/wiki/spaces/ENG/pages/1283916646/How+to+burn+and+validate+prewarmed+freeipa+image+for+testing+purposes

lacikaaa (approved):

handavid:
@lacikaaa thanks for the info! I kicked off the jenkins job

handavid:
@lacikaaa I ran the job but validation failed. It doesn't look like the failure is related to my change though since the environment reached the AVAILABLE state. There are also a few similar errors sprinkled through the history. Any ideas on how to overcome this?

```
com.sequenceiq.it.cloudbreak.exception.TestFailException: Cluster 'aws-test-dff1ac1482' entered into creation failed state. Status: '{status=PROVISIONING_FAILED}' statusReason: '{statusReason=Datalake creation failed. Cannot start provisioning, error happened during the operation: Error message: "Image id settings are incorrect: java.lang.RuntimeException: com.sequenceiq.cloudbreak.core.CloudbreakImageNotFoundException: Could not find any image for platform 'aws', runtime '7.2.13' and Cloudbreak version '2.52.0-b104' in 'v3-dev' image catalog."}'
```
http://ci-cloudbreak.eng.hortonworks.com/view/e2e-test/job/api-e2e-imagevalidator-aws/1817/

pkedvessy:
@handavid We are updating the default runtime what's going to fix the issue. We are going to re-run the validation as well once the default runtime change is merged and available on mow-dev.

handavid:
@pkedvessy thanks for the update!

handavid:
@lacikaaa I did some further testing on the behavior of the parameters added to the disable and delete commands. I had thought that passing the parameters in was harmless for unpatched FreeIPAs but this is not the case. There is no error returned, but the user is not actually deleted from FreeIPA.

This means that either:
  1. the user_disable and user_del commands should not have the flag or
  2. the FMS should be aware of whether FreeIPA is patched so that it can send or not send the flag

I'm going to take that change out of this PR and open a separate JIRA to track that work.

lacikaaa:
the change is LGTM by me. I don't know if we have to rerun the image validation stuff. @pkedvessy

pkedvessy:
There was no succeeded verification yet, so yes we should.

pkedvessy:
@handavid Validation passed: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-aws/1826/

PR #665, opened 2022-02-08
CB-15128 Set nologin shell for postgresql

Bajzathd:
Can only be merged after https://github.com/hortonworks/cloudbreak/pull/12235 reaches prod to avoid issues

### DIFF
--- a/saltstack/base/salt/postgresql/init.sls
+++ b/saltstack/base/salt/postgresql/init.sls
@@ -146,3 +146,8 @@ configure-listen-address:
 {% if pillar['subtype'] != 'Docker' %}
       - service: start-postgresql
 {% endif %}
+
+set-postgres-nologin-shell:
+  user.present:
+    - name: postgres
+    - shell: /usr/sbin/nologin


saltstack/base/salt/prerequisites/user.sls
```diff
@@ -1,3 +1,9 @@
+{% if salt['environ.get']('OS').startswith("centos") %}
+centos_user_noshell:
+  cmd.run:
```
lacikaaa, line 3: https://docs.saltproject.io/en/latest/ref/states/all/salt.states.user.html
user.present.shell would be a nicer way to set this

saltstack/base/salt/prerequisites/user.sls
```diff
@@ -1,3 +1,9 @@
+{% if salt['environ.get']('OS').startswith("centos") %}
```
lacikaaa, line 1: os grain could be used for this I think, like here: https://github.com/hortonworks/cloudbreak/blob/ffde9cb187d09da8d3964510bb634cf04cab0f83/orchestrator-salt/src/main/resources/salt/salt/metering/init.sls#L10

Bajzathd:
@lacikaaa Thanks for the comments, but the centos user modification is already merged
https://github.com/hortonworks/cloudbreak-images/pull/666/files
I will open a separate PR with your improvement ideas

saltstack/base/salt/postgresql/init.sls
```diff
@@ -146,3 +146,7 @@ configure-listen-address:
 {% if pillar['subtype'] != 'Docker' %}
       - service: start-postgresql
 {% endif %}
+
+postgres:
+  user.present:
+    - shell: /usr/sbin/nologin
```
lacikaaa, line 152: could you give a meaningful state name instead of using the user as the state name? eg:
```
set-postgres-login-shell:
  user.present:
    - name: postgres
    - ...
```

lacikaaa (approved):

pkedvessy (approved):

PR #670, opened 2022-02-10
CB-15727 Remove old kernels

Bajzathd:
To be able to remove old kernels after system update, the machine has to be rebooted. To achieve this, the final saltstate was extracted to a new salt-final.sh script, and a reboot is issued before running it.

### DIFF
--- a/saltstack/final/salt/cleanup/user.sls
+++ b/saltstack/final/salt/cleanup/user.sls
@@ -1,0 +1,8 @@
+{% if grains['os'] == 'CentOS' %}
+
+set_centos_nologin_shell:
+  user.present:
+    - name: centos
+    - shell: /usr/sbin/nologin
+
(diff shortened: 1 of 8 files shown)


saltstack/final/salt/cleanup/user.sls
```diff
@@ -0,0 +1,7 @@
+{% if grains['os'] == 'CentOS' %}
+
+centos:
```
lacikaaa, line 3: similarly to the other pr, I prefer states with names to avoid state name collision. please give a proper name to the state

lacikaaa (approved):

saltstack/final/salt/cleanup/user.sls
Bajzathd: Fixed, thanks

PR #681, opened 2022-03-22
CB-16538: Burn prewarmed images with JDK11 installed alongside JDK8

jszeles:
We need JDK 11 installed alongside JDK8 on our prewarmed images

### DIFF
--- a/saltstack/hortonworks/pillar/java/init.sls
+++ b/saltstack/hortonworks/pillar/java/init.sls
@@ -29,9 +29,15 @@ openjdk_packages:
 {% elif salt['environ.get']('OS') == 'redhat7' %}
   - java-1.8.0-openjdk-headless
   - java-1.8.0-openjdk-devel
+  - java-11-openjdk-headless
+  - java-11-openjdk-devel
 {% else %}
   - java-1.8.0-openjdk-headless
   - java-1.8.0-openjdk-devel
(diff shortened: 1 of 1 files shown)


saltstack/hortonworks/pillar/java/init.sls
```diff
@@ -29,9 +29,15 @@ openjdk_packages:
...
+  - java-11-openjdk-devel
 {% else %}
   - java-1.8.0-openjdk-headless
   - java-1.8.0-openjdk-devel
   - java-1.8.0-openjdk-javadoc
   - java-1.8.0-openjdk-src
+  - java-11-openjdk-headless
+  - java-11-openjdk-devel
+  - java-11-openjdk-javadoc
+  - java-11-openjdk-src
```
Bajzathd, line 42: Which java version will be the default if you install both?

jszeles: The default is jdk 8, we checked in a running instance.

TheTinkerDad (approved):

saltstack/hortonworks/pillar/java/init.sls
pkedvessy: I'd more prefer to force Java8 to be the default, instead of relying on it was default on a random running instance...

PR #693, opened 2022-04-27
CB-16947 Apply AES256 in decrypting key and SHA512 for HMAC check.

bergerdenes:
Take IV as parameter instead of fixed value.
If the HMAC key was provided, then the HMAC digest is
expected, the latter is computed in-place as well
and is compared against the received digest.

### DIFF
--- a/saltstack/base/salt/prerequisites/usr/bin/ccmv2-helper.sh
+++ b/saltstack/base/salt/prerequisites/usr/bin/ccmv2-helper.sh
@@ -12,11 +12,11 @@ setup_ccmv2() {
 
   mkdir -p /etc/ccmv2
 
-  IV=436c6f7564657261436c6f7564657261
+  LEGACY_IV=436c6f7564657261436c6f7564657261
   AGENT_KEY_PATH=/etc/ccmv2/ccmv2-key.enc
 
   CCM_V2_AGENT_KEY_HEX=$(xxd -pu -l 16 <<< $CCM_V2_AGENT_KEY_ID)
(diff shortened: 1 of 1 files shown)


saltstack/base/salt/prerequisites/usr/bin/ccmv2-helper.sh
```diff
@@ -12,11 +12,11 @@ setup_ccmv2() {
 
   mkdir -p /etc/ccmv2
 
-  IV=436c6f7564657261436c6f7564657261
+  LEGACY_IV=436c6f7564657261436c6f7564657261
   AGENT_KEY_PATH=/etc/ccmv2/ccmv2-key.enc
 
   CCM_V2_AGENT_KEY_HEX=$(xxd -pu -l 16 <<< $CCM_V2_AGENT_KEY_ID)
-  echo ${CCM_V2_AGENT_ENCIPHERED_KEY} | openssl enc -aes-128-cbc -d -A -a -K ${CCM_V2_AGENT_KEY_HEX} -iv ${IV} > ${AGENT_KEY_PATH}
+  echo ${CCM_V2_AGENT_ENCIPHERED_KEY} | openssl enc -aes-128-cbc -d -A -a -K ${CCM_V2_AGENT_KEY_HEX} -iv ${LEGACY_IV} > ${AGENT_KEY_PATH}
```
SachinHg, line 19: There should be an equivalent of this when we use aes-256 for encryption as well.

bergerdenes: well, it works (tested both cases)

SachinHg: Ok. As clarified by Denes over slack, AGENT_KEY_PATH is required in case of mTLS only.

SachinHg (approved):
lgtm

TheTinkerDad (approved):

PR #696, opened 2022-05-12
CDPAM-3145: DataLake Creation Failed

zabhi:
To reach out to https endpoints, golang depends on the value in `https_proxy` variable. Earlier, golang used to fall back to `http_proxy` variable inn the absence of an `https_proxy` variable.
After a recent golang version update (1.16), the behavior was changed to stop the client falling back to `http_proxy` variable. This broke our NTP proxy support.

This change fixes the problem by setting the correct variable.


saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -182,7 +182,7 @@ setup_proxy() {
     fi
     PROXY_ENV_FILE=/etc/cdp/proxy.env
     mkdir -p /etc/cdp
-    echo ${PROXY_PROTOCOL}_proxy=$PROXY_URL > $PROXY_ENV_FILE
+    echo https_proxy=$PROXY_URL > $PROXY_ENV_FILE
```
SachinHg, line 185: what about if the proxy is of type `http_proxy`?

lacikaaa: yes this seems odd, I would expect to use the correct protocol. maybe we should set both? is there any automated tests for these?

zabhi: The variable name is tied to the target URL and not the proxy' protocol. So, the original code was incorrect.

Earlier, golang used to fall back to `http_proxy` variable in the absence of an `https_proxy` variable.
After a recent golang version update (1.16), the behavior was changed to stop the client falling back to `http_proxy` variable. This broke our NTP proxy support. (See https://go.dev/doc/go1.16#library)

So, if the client needs to connect to `https://example.com`, the variable used would be `https_proxy`, even if the proxy address uses http (eg. `http://proxy.host:5138`)

lacikaaa (changes requested):
please squash your commits

lacikaaa (approved):

SachinHg (approved):
lgtm

meetrajsinhvala:
We have baked freeipa images from this branch and below are the image-UUIDs for different cloud providers.

- AWS : 90fa3754-121d-4b08-88cd-39d7cc4080e2
- Azure : bb13bbe6-8afc-454f-b370-aecb05cbb9fa 
- GCP : caf361ec-1089-4a4c-ab41-ca6e77dd95c6

We have tested this with our CCMv2 dev test suite and CCMv2 e2e test suite (with proxies). CCMv2 dev test runs and e2e tests run were successful.

PR #697, opened 2022-05-16
CB-17072 - increase the salt retry count upon unsuccessful attempt

gregito:
This PR is due to the following escalation: ENGESC-14373


lacikaaa (changes requested):
could you provide a bit more elaborate commit message, like:
- what was the previous setting
- what are we trying to solve here
- why are these values chosen

lacikaaa (approved):

TheTinkerDad (approved):

PR #704, opened 2022-06-21
CB-17500 Install prometheus on base image

foldik:
Install prometheus that we will use for compute monitoring.

Jenkins builds:
- http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/6028/console
- http://build.eng.hortonworks.com/job/imagecatalog-v3-append/4360/console

I only changed the name of the package in the `/tmp/package-versions.json` from prometheus to cdp-prometheus.

### DIFF
--- a/saltstack/base/salt/monitoring/cdp-prometheus.sls
+++ b/saltstack/base/salt/monitoring/cdp-prometheus.sls
@@ -1,0 +1,15 @@
+/opt/cdp-prometheus:
+  file.directory:
+    - name: /opt/cdp-prometheus
+    - user: root
+    - group: root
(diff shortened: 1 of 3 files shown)


Bajzathd (approved):
Nit: what makes it cdp-prometheus? I think it is just prometheus. Also, other directories under /opt has underscores in their names instead of dashes.

oleewere:
@Bajzathd,that's common tool to be installed, so this way it is in a specific places + we excluded promtool from the bin to make it smaller, as it will be used only agent mode by us (so it will be specifically used by us)

PR #705, opened 2022-06-22
CB-17407 Install postgresql 11 next to 10 on the images

lacikaaa:
The default is still postgresql 10. Both databases are initialized.
Tested with creating AWS and docker image.


saltstack/base/salt/postgresql/init.sls
```diff
@@ -20,6 +28,121 @@ install-postgres:
       - postgresql10-docs
       - postgresql10-devel
 
+
+install-centos-sclo-rh:
```
Bajzathd, line 32: typo

saltstack/base/salt/postgresql/init.sls
```diff
@@ -20,6 +28,121 @@ install-postgres:
...
+  cmd.run:
+    - name: yum -y install centos-release-scl-rh
```
Bajzathd, line 34: I think this should be replaced with pkg.installed

lacikaaa: right, let me try it out first

Bajzathd:
Do we need to initialize and run both? In case of java11 only the package was installed, and the initialization was done with recipes for testing. Maybe the same could be done here @pkedvessy ?
Also, could this cause any performance issues to have 2 dbs running?

Bajzathd (approved):

PR #706, opened 2022-07-06
CB-17502. Add json and open metrics support for cipa script

oleewere:


lacikaaa (changes requested):
could you add a bit more description to the commit message? why is it necessary, what is changed, etc

lacikaaa (approved):

pkedvessy (approved):

PR #714, opened 2022-08-19
CB-18203: Implement a script that generates individual package changelogs

jszeles:
We need a simple script that iterates through all the RPM packages of the OS and using "rpm -q --changelog" it generates a .txt file for each package in a separate folder for later processing. This script will have to be executed on VMs started with our Cloudera Runtime or FreeIPA images by Packer.

### DIFF
--- a/scripts/changelog/collect-package-changelogs.sh
+++ b/scripts/changelog/collect-package-changelogs.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+set -e
+[[ $DEBUG ]] && set -x
+
+if [ -z "${IMAGE_UUID}" ] ; then
+  echo "IMAGE_UUID env variable is mandatory"
+  exit 1
+fi
+
+BASE_PATH="/tmp/changlogs-tmp"
+RPM_PACKAGE_LIST_PATH="${BASE_PATH}/rpm-packages.txt"
+RPM_PACKAGE_TAR_FILE_NAME="rpm-package-changelogs.tar.gz"
+RPM_PACKAGE_TAR_FILE_PATH="${BASE_PATH}/${RPM_PACKAGE_TAR_FILE_NAME}"
+CHANGELOG_DIRECTORY_NAME="${IMAGE_UUID}-changelogs"
+CHANGELOG_BASE_PATH="${BASE_PATH}/${CHANGELOG_DIRECTORY_NAME}"
+
+FILTERED_RPM_PACKAGES=("gpg-pubkey")
+
+function execute() {
+  echo "Cleaning up the temporary folders and files..."
+  clean_up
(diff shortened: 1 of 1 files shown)


scripts/changelog/collect-package-changelogs.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+
+set -e
+[[ $DEBUG ]] && set -x
+
+if [ -z "${IMAGE_UUID}" ] ; then
+  echo "IMAGE_UUID env variable is mandatory"
+  exit 0
```
pkedvessy, line 8: It seems to be error case so return code should not be 0.

scripts/changelog/collect-package-changelogs.sh
```diff
@@ -0,0 +1,99 @@
...
+  echo "Storing changelogs to file by rpm package name..."
+  store_rpm_package_changelog_to_file
+  echo "Compressing the entire directory that contains generated files into ${RPM_PACKAGE_TAR_FILE_PATH} file..."
+  cd $BASE_PATH
+  tar -zcf $RPM_PACKAGE_TAR_FILE_NAME $CHANGELOG_DIRECTORY_NAME
+}
+
+function clean_up() {
+  if [ -f $RPM_PACKAGE_LIST_PATH ] ; then
+    rm $RPM_PACKAGE_LIST_PATH
```
pkedvessy, line 34: You can simplify file removes by using rm with -f flag so you won't need the check the file existence.
Or what I would more prefer: you can use a specific folder inside /tmp as BASE_PATH so you can simply delete the BASE_PATH folder itself.

scripts/changelog/collect-package-changelogs.sh
```diff
@@ -0,0 +1,99 @@
...
+  amazon)
+    execute
+    ;;
+  suse)
+    echo "Platform does not support this functionality"
+    exit 0
+    ;;
+  *)
+    echo "Unsupported platform:" $SALT_INSTALL_OS
+    exit 0
```
pkedvessy, line 97: It seems to be error case so return code should not be 0.

scripts/changelog/collect-package-changelogs.sh
jszeles: done

scripts/changelog/collect-package-changelogs.sh
jszeles: done

scripts/changelog/collect-package-changelogs.sh
jszeles: new specific folder introduced for storing all temporary files

pkedvessy (approved):

PR #718, opened 2022-08-31
CB-18227: Changed SSHD ClientAliveInterval to 235

TheTinkerDad:
Basically this is the highest value allowed by Azure Marketplace,
however CB will change this to 1800 during instance provisioning,
because a much larger value is essential to our own image validation
and E2E processes.

**Also DON'T MERGE this before Cloudbreak 2.62.0-b84 reaches production
or you unleash Ragnarok on us by crippling image releases!**

Warned you...

### DIFF
--- a/saltstack/final/salt/cis-controls/init.sls
+++ b/saltstack/final/salt/cis-controls/init.sls
@@ -95,12 +95,14 @@ sshd_harden_addressLoginGraceTime:
     - repl: "LoginGraceTime 60"
     - append_if_not_found: True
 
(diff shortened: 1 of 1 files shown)


Bajzathd (approved):

TheTinkerDad:
According to information gathered lately (thanks @lajosrodek for the hint!), nothing besides our end to end tests rely on SSH. Given our Azure validation test suite completed successfully with this change active, I assume it's safe to merge this. In worst case we'll have to revert...

PR #726, opened 2022-09-23
CB-18202: Increase packer version due to the skip create ami feature

jszeles:
Now we use packer version 1.4.2 that cannot support the skip create ami feature, so we should increase the version number >= 1.7.0 where this feature was provided at the first time. The latest version will be tried to used.


sgabor001 (approved):
LGTM

PR #731, opened 2022-10-05
CB-18746 Increase Azure Centos7 image size as image burnings are failing with no space left on device

pkedvessy:


TheTinkerDad (approved):

TheTinkerDad:
Approved this magical piece of hack, but we should re-test it from branch because we've changed a lot in the repo recently...

PR #735, opened 2022-10-10
RHEL8 FreeIPA on AWS and AWS_GOV

pkedvessy:


saltstack/base/salt/prerequisites/jinja.sls
```diff
@@ -1,3 +1,9 @@
+{% if pillar['OS'] == 'redhat8' %}  
 install_jinja2:
   cmd.run:
```
TheTinkerDad, line 3: You can use this instead:

```
install_jinja2:
  pip.installed:
      - name: jinja2
```

pip3 is not always available or on the path - the above solution takes the pip available for Salt and uses that one - usually the one that is bound to the virtualenv Salt is running in. With that, you don't need the "if" to differentiate between OSes.

saltstack/base/salt/prerequisites/pip.sls
```diff
@@ -24,15 +24,21 @@ update_python_pip2:
...
-    - onlyif: pip3 -V
+    - name: python3 -m pip install --upgrade --index=https://pypi.python.org/simple/ pip==8.1.2
+    - onlyif: python3 -m pip -V
 
 {% endif %}
 
+{% if pillar['OS'] == 'redhat8' %}  
+install_pyyaml:
+  cmd.run:
+    - name: python3 -m pip install "PyYAML>=5.1" --ignore-installed
```
TheTinkerDad, line 35: Suggestion:

```
install_pyyaml:
  pip.installed:
    - name: PyYAML
```
    
Also, we're installing PyYAML 6.0 by default either way - you can get rid of the whole "if" part too

saltstack/freeipa/salt/ipaconsistency/init.sls
```diff
@@ -1,6 +1,6 @@
 install_checkipaconsistency:
   cmd.run:
-    - name: pip install checkipaconsistency==2.7.10
+    - name: python3 -m pip install checkipaconsistency==2.7.10
```
TheTinkerDad, line 3: Suggestion:

```
install_checkipaconsistency:
  pip.installed:
    - name: checkipaconsistency==2.7.10
```

scripts/generate-delta-package-list.sh
```diff
@@ -73,7 +73,11 @@ function collect_rpm_packages() {
 function collect_python_packages() {
   file=$1
   echo "Collecting installed python packages"
-  pip list --format json | jq -r '.[].name' | sort >> "$file"
+  if [ "${OS}" == "redhat8" ] ; then
```
TheTinkerDad, line 76: You could replace "virtualenv" and "pip" with "python3 -m virtualenv" and "python3 -m pip" in general - we shouldn't rely on anything Python 2 related nowadays, not even on CentOS 7.

scripts/get-salt-version.sh
```diff
@@ -30,7 +30,7 @@ compare_version () {
 BASE_NAME=$1
 STACK_VERSION=$2
 
-SALT_VERSION=3001.8
+SALT_VERSION=3004.1
```
TheTinkerDad, line 33: Isn't 3004.1 something we previously couldn't update to? Or is this an RHEL / Gov related mandatory change? Personally I wouldn't do this together with all of the changes required for Gov, but instead separately under https://jira.cloudera.com/browse/CB-16854 on master and test it with CentOS 7 first.

scripts/salt-install.sh
```diff
@@ -20,14 +20,30 @@ function install_salt_with_pip() {
...
     pip install pbr
   fi
   pip install --upgrade pip
   pip install -r /tmp/salt_requirements.txt
 }
 
+function install_salt_with_pip3() {
+
+  echo "Installing salt with version: $SALT_VERSION"
+  pip3 install --upgrade pip
```
TheTinkerDad, line 34: At some places you've removed "pip3" and replaced it with "python3 -m pip". Why do you stick to "pip3" here?

TheTinkerDad (commented):
I'm not requesting changes, only suggested a few - I'll leave it up to you whether you want to adopt them or not. Otherwise, LGTM

TheTinkerDad (approved):

scripts/get-salt-version.sh
pkedvessy: Agree, it came in dunno why... Maybe it was there by a mistake on the base branch I used.

saltstack/base/salt/prerequisites/jinja.sls
TheTinkerDad: Already done on another branch

saltstack/freeipa/salt/ipaconsistency/init.sls
TheTinkerDad: Already done on another branch

PR #738, opened 2022-10-17
CB-18661 avoid the "ssl" directive is deprecated warning message, and…

akanto:
… delete the dead code and free up the port 1080.

### DIFF
--- a/README.image-content.md
+++ b/README.image-content.md
@@ -15,11 +15,11 @@ This section provides information about the packages of specified images to be i
 
 ## Source of base images
 
- Cloud Provider | OS | Source
- ---- | ---- | ----
- AWS | CentOS 7 | `ami-098f55b4287a885ba`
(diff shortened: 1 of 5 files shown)


akanto:
This is more than 3 months old

TheTinkerDad:
> This is more than 3 months old

Yes, and we still couldn't decide whether it was some kind of experiment, an unfinished PR, a leftover PoC, etc. No proof of test (image burned from branch), no ticket, no description. Whatever we merge in we do it super carefully because it can have an immediate impact on images we send to production or it can simply hinder our daily operation.

akanto:
I see, the counterpart of this where we are overwriting the ssl.conf is in production since October: https://github.com/hortonworks/cloudbreak/commit/07e4c37a9992fb44366b73f5dfa8c97be1c9bd17

I have tested it manually overwriting the mentioned files.

TheTinkerDad (approved):

PR #752, opened 2022-12-01
CB-19843 - Images not published in ap-east-1 region (Hong Kong)

sgabor001:
After copying the image, we need to check the status of the image in order to give feedback to the jenkins job.

### DIFF
--- a/scripts/aws-copy.sh
+++ b/scripts/aws-copy.sh
@@ -49,6 +49,20 @@ function wait_for_image() {
   log "Copy operation in region $REGION for image $AMI_IN_REGION finished"
 }
 
+function check_if_image_is_public_and_available_in_region() {
+  REGION=$1
+  AMI_IN_REGION=$2
+  IMAGE_DESC=$(aws ec2 describe-images --region $REGION --image-ids $AMI_IN_REGION)
+  REGEX_PUBLIC="Public: true"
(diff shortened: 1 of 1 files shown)


scripts/aws-copy.sh
```diff
@@ -49,6 +49,20 @@ function wait_for_image() {
...
+function check_if_image_is_public_and_available_in_region() {
+  REGION=$1
+  AMI_IN_REGION=$2
+  IMAGE_DESC=$(aws ec2 describe-images --region $REGION --image-ids $AMI_IN_REGION)
+  REGEX_PUBLIC="Public: true"
+  REGEX_STATE="State: available"
+  if [[ $IMAGE_DESC =~ $REGEX_PUBLIC ]] && [[ $IMAGE_DESC =~ $REGEX_STATE ]]; then
+    echo "The $AMI_IN_REGION is PUBLIC and AVAILABLE in $REGION region."
+  else
+    echo "FAILURE | The $AMI_IN_REGION in $REGION region is not available or not in correct state."
```
Bajzathd, line 61: I think separate error messages would be useful for not available and not public images

Bajzathd: Also, please use `log` instead of `echo` to include the current timestamp

Bajzathd (approved):

PR #757, opened 2023-01-03
CB-19582 CB-19583 Configure SSH/user session timeout for all systems

lacikaaa:
Currently the timeout was added only for Centos 7. Refactored the related states:
- it's applied to all OS
- matches the structure of how we configure profile


lacikaaa:
tested with centos7 and rh8

schfeca75 (approved):

TheTinkerDad (approved):

PR #759, opened 2023-01-06
CB-19581 Gov cloud compatible SSH banner

lacikaaa:
In case of gov cloud image replace the current text after login.
- moved the config from cis to the common configuration so it's added for every OS
- `-` aroung `%` is to ensure no blank lines are inserted
- `indent(8)` to ensure the rendered state file is valid, without formatting the text itself


lacikaaa:
tested with non gov centos7 and gov rh8

pkedvessy (approved):

PR #763, opened 2023-01-18
CB-17441 Set the domain in case of Java11 when making resolv.conf

pkedvessy:


TheTinkerDad (approved):

lacikaaa:
you missed to add to the commit message why is this change required

pkedvessy:
@lacikaaa  True, thanks for calling it out. I missed it because it is coming from a different branch by a cherry-pick...

We need the domain in the resolv.conf to have java.net.InetAddress.getHostName() consistent in Java8 and Java11.
In case of Java8 it returns the FQDN but without the domain in the resolv.conf Java11 returns only the hostname and at the end it resoults in a DNS resolution issue in CM: "The hostname and canonical name for this host are not consistent when checked from a Java process."

I'm going to add this message to the commit.

saltstack/hortonworks/salt/dhcp/etc/dhcp/dhclient-enter-hooks
```diff
@@ -21,6 +21,40 @@ log() {
...
+
+    echo "nameserver 127.0.0.1" >> /etc/resolv.conf
+
+    if [ ! -f /etc/hostname ]; then
+        log "Creating new hostname"
+        echo $new_host_name > /etc/hostname
+        hostname -F /etc/hostname
+    fi
+
+    JAVA_VERSION=$(java -version 2>&1 | grep -oP 'version [^0-9]?(1\.)?\K\d+' || true)
```
lacikaaa, line 45: while this could work it would be more simple solution to use salt templating to generate the script during image burning instead of running this check every time some dhcp event happens

pkedvessy: Its a bit more complicated if we change the default java version at provision time what the plan is as a part of CRB-2312

lacikaaa: okay, thanks for the clarification

PR #772, opened 2023-02-02
CB-20785: Fix global PATH for Python 3.8 on CentOS7

TheTinkerDad:


TheTinkerDad:
Test image burning - /job/cloudbreak-multi-packer-image-builder-salt-v3/7716/

Bajzathd (approved):

PR #777, opened 2023-02-10
CB-19862 Do not create compute images on GCP for 7.2.17 runtime

Bajzathd:
Skip creating compute images on GCP as we are only using their tar.gz counterpart.


Bajzathd:
Test image with successful validation: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-gcp/1188/

pkedvessy (approved):

PR #785, opened 2023-02-27
CB-20904: Disabled Python 3.8 for RT < 7.2.17 and CentOS 7

TheTinkerDad:


scripts/salt-install.sh
```diff
@@ -109,9 +109,17 @@ function install_python_pip() {
       yum install -y python36 python36-pip python36-devel python36-setuptools
     fi
 
-  # For images with Runtime 7.2.15 and below we only support RHEL7 and CentOS7 with Python 2.7 and 3.6
-  elif [ $(version $STACK_VERSION) -le $(version "7.2.15") ]; then
-    if [ "${OS}" == "redhat7" ] ; then
+  # For images with Runtime 7.2.16 or lower:
+  #   - RHEL7 and CentOS7 stays with Python 2.7 and 3.6
+  #   - RHEL8 goes with Python 3.8 (Runtime <= 7.2.15 won't be officially supported though)
+  elif [ $(version $STACK_VERSION) -le $(version "7.2.16") ]; then
```
Bajzathd, line 115: I think this would make more sense to move into the original if condition (on line 150 currently). It would also eliminate the redhat8 python install duplication.

TheTinkerDad: Left it separated for readability's sake - this is a temporary rollback and will be reverted eventually (with 7.2.16.2), it'll be easier to revert this way.

TheTinkerDad:
Tested from branch. Going to merge it, because currently this is a blocker for 7.2.16.1 and JDK11 opt-in and we need the new images tonight.

PR #793, opened 2023-03-21
CB-20186 [FedRamp] | Skip rpms on aws gov installation

sgabor001:


sgabor001:
Originally it was similar but with extra conditions regarding RHEL8 and errors occured. I am not familar with salt but seems there are limitations on complex conditions. That is why I brought it to outside and inside remained the variable equal check as in other places.

I tested it here:
http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/207/consoleFull
based on the logs the minifi and request-signer were not installed.

PR #796, opened 2023-03-24
CB-21044 Patch FreeIPA constants lib to support hostnames longer than…

lacikaaa:
… 64 chars

RedHat claims that hostname must be the same as FQDN thus have the limit 64 char for the hostname/FQDN. As we don't set the hostname to FQDN, it's possible the FQDN is longer than 64 char, but this limit would cause failures on client side when `ipa-client-install` is run. This change is introduced in RH8, Centos7 works fine. Commit causing the validation error: https://github.com/freeipa/freeipa/commit/6662e99e173a16059271528e1fa58e37a3028924

The limit on FreeIPA server side is also increased: https://github.com/hortonworks/cloudbreak/commit/18a867c6b47fc6efb39dd50c3f5557f3940f25bb

### DIFF
--- a/saltstack/base/salt/prerequisites/init.sls
+++ b/saltstack/base/salt/prerequisites/init.sls
@@ -24,6 +24,8 @@ include:
   - {{ slspath }}.disable-unattended-upgrades
 {% endif %}
   - {{ slspath }}.authconfig
+  - {{ slspath }}.ipa
+
 
 /usr/bin/:
   file.recurse:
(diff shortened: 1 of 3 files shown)


lacikaaa:
@akanto I've replaced `file.replace` with `file.patch` which would fail if the patch could not be done for any reason. Tested with moving the line and deleting it

akanto (approved):

pkedvessy (approved):

PR #817, opened 2023-05-23
CB-21931 forward Shared Address Space addresses from unbound

sodre90:


lacikaaa (commented):
why don't you just get rid of these local zones and switch on `unblock-lan-zones`?

biharitomi (approved):

Bajzathd (approved):

PR #829, opened 2023-06-12
CB-21583 CIS: Use tmpfs for /tmp

Bajzathd:


saltstack/final/salt/cis-controls/init.sls
```diff
@@ -558,6 +558,13 @@ dev_shm_noexec:
 dev_shm_remount:
   cmd.run:
     - name: 'sudo mount -o remount,noexec,nodev,nosuid /dev/shm'
+#1.1.2 Ensure /tmp is configured
```
TheTinkerDad, line 561: @lajosrodek @lacikaaa I'd like to ask for your opinion here. Setting this up IS required for CIS compliance, however I'm a bit concerned about performance implications. By default the tmpfs can eat up as much as half of the available RAM, not considering swap, possibly forcing the OS to make some services to rely on swap. Do you consider this dangerous? If so, how would you reduce the chance for problems? I'm thinking about something like limiting tmpfs max size to only a few gigs, etc.

lacikaaa: we definitely should limit the size.  The issue might be that we have to decide the size during image burning, while a wide range of instance types could be used. So I would check the smallest one which is used, and try to size it based on that. I don't think we can allow more than 2-3GB. The question is if it's enough

lacikaaa: side note: it's your territory, but I'm not a fan of this single cis file. While it keeps everything cis related in one place, most of the settings would fit into already existing states. Eg I would think mount or storage (I don't know why the latter exists) be a better place for this, as I would expect all filesystem mount logic to be placed there.

Bajzathd: > The issue might be that we have to decide the size during image burning

We could also move the sizing step to the user data script

> I would check the smallest one which is used

I've found we support AWS c1.medium instance type which has 1,7GB memory and also c5.24xlarge with 192GB memory so a fixed value does not seem reasonable, maybe a fixed percentage lower than the default 50%?

Bajzathd: @lacikaaa we discussed with @TheTinkerDad that 30% would seem sufficient, do you agree?

lacikaaa: we could try, as I think we don't have any data on prod memory usage ATM, but we might want to notify QE about this change

lajosrodek:
Shall we close this since the ticket has been refused as Won't FIx?

PR #832, opened 2023-06-12
CB-22127: Temporary fix for the broken cdp-logging-agent (v2)

TheTinkerDad:
For CentOS 7 the new version is optional, so disabled it. For RHEL 8 w/ FIPS, it's mandatory, so it's still a blocker bug and a new RPM will be needed. Had to re-write this due to the weird RPM selection logic...

### DIFF
--- a/saltstack/base/salt/telemetry/centos/init.sls
+++ b/saltstack/base/salt/telemetry/centos/init.sls
@@ -1,7 +1,7 @@
 {% set include_cdp_telemetry = salt['environ.get']('INCLUDE_CDP_TELEMETRY') %}
 {% set include_fluent = salt['environ.get']('INCLUDE_FLUENT') %}
-{% set cdp_telemetry_rpm_repo_url = salt['environ.get']('CDP_TELEMETRY_RPM_URL') %}
-{% set cdp_logging_agent_rpm_repo_url = salt['environ.get']('CDP_LOGGING_AGENT_RPM_URL') %}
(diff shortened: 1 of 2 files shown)


sgabor001 (approved):

saltstack/base/salt/telemetry/centos/init.sls
```diff
@@ -33,14 +33,14 @@ install_cdp_telemetry_rpm:
...
-{% elif cdp_logging_agent_rpm_repo_url %}
+{% if cdp_logging_agent_rpm_url %}
 install_cdp_logging_rpm_from_repo_url:
   cmd.run:
-    - name: rpm -i {{ cdp_logging_agent_rpm_repo_url }}
+    - name: rpm -i {{ cdp_logging_agent_rpm_url }}
+{% elif use_telemetry_archive == "Yes" %}
+install_cdp_logging_rpm_from_archive:
+  cmd.run:
+   - name: "/cdp/telemetry/install-package.sh cdp_logging_agent {{ archive_base_url }} {{ archive_credentials }} {{ cdp_logging_agent_rpm_url }}"
```
pkedvessy, line 43: cdp_logging_agent_rpm_url will be empty here beause of the if confition in the line 39 . Do I miss something?

TheTinkerDad: Yes, that's a harmless side effect of switching up the "if"s.

Since I plan to open a separate ticket for cleaning up this mess anyway, I left it that way.

pkedvessy (approved):

PR #847, opened 2023-06-22
CDPCP-9042: Added the rewritten heartbeat application for AWS Gov

TheTinkerDad:


sgabor001 (approved):

pkedvessy (changes requested):
Seems like the rpm name change triggers CB to install the snapshot version of the original rpm

PR #850, opened 2023-06-28
CB-12687: Copy image to all regions as part of prod promotion

jszeles:
Only prod catalog needs all regions of an image. Before that we should only store the images in a few regions where testing takes place.

We should copy the images only to us-west-2 and eu-central after validation, like it is now. The rest of the copy should happen right before prod promotion. If a complaint arrives that someone needs images in other regions as well for testing/development/etc. and the request is justified (e.g. other regions are slow or something) we'll re-enable early copy for those regions as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,11 +5,13 @@ ATLAS_PROJECT ?= "cloudbreak"
 ENABLE_POSTPROCESSORS ?= ""
 CUSTOM_IMAGE_TYPE ?= "hortonworks"
 IMAGE_OWNER ?= "cloudbreak-dev"
-#for oracle JDK use oracle-java
+# for oracle JDK use oracle-java
 OPTIONAL_STATES ?= ""
 # only for oracle JDK
 ORACLE_JDK8_URL_RPM ?= ""
 SLES_REGISTRATION_CODE ?= "73D5EBB68CB348"
+# for splitting image copy (test and prod phases)
+IMAGE_COPY_PHASE ?= ""
 
 # Azure VM image specifications
 ifeq ($(CLOUD_PROVIDER),Azure)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -162,24 +164,36 @@ endif
...
-asia-east1,asia-east2,asia-south2,asia-southeast1,australia-southeast1,australia-southeast2,asia-northeast3,europe-west1,europe-west2,europe-west3,europe-west4,us-west1,us-west2,us-west3,europe-west6,us-west4,southamerica-east1,us-central1,europe-north1,europe-central2,northamerica-northeast1,northamerica-northeast2,us-east4,asia-south1,us-east1,asia-northeast2,asia-northeast1,asia-southeast2,southamerica-west1
+# AWS AMI region definitions for the different states
+define AWS_AMI_REGIONS_PHASE1
+us-west-1,us-west-2,eu-central-1
 endef
 
-ifndef AWS_AMI_REGIONS
-define AWS_AMI_REGIONS
-ap-northeast-1,ap-northeast-2,ap-south-1,ap-southeast-1,ap-southeast-2,ap-southeast-3,ca-central-1,eu-central-1,eu-west-1,eu-west-2,eu-west-3,sa-east-1,us-east-1,us-east-2,us-west-1,us-west-2,eu-north-1,eu-south-1,af-south-1,me-south-1,ap-east-1,eu-south-2,eu-central-2,me-central-1
+define AWS_AMI_REGIONS_PHASE2
```
Bajzathd, line 172: Copy jobs should handle when an image is already copied to a region, so I would simply use the original `AWS_AMI_REGIONS`, `AWS_GOV_AMI_REGIONS` and `AZURE_STORAGE_ACCOUNTS` lists as phase 2
This has 2 benefits:

1. We have the list of all regions in a single place
2. It would make sure that we have the image in the phase 1 regions

Makefile
```diff
@@ -214,6 +226,31 @@ Switzerland North:cldrswitzerlandnorth,\
...
+	AZURE_STORAGE_ACCOUNTS=$(AZURE_STORAGE_ACCOUNTS_PHASE2)
+endif
+endif
+ifndef AZURE_STORAGE_ACCOUNTS
+	AZURE_STORAGE_ACCOUNTS=$(AZURE_STORAGE_ACCOUNTS_PHASE1),$(AZURE_STORAGE_ACCOUNTS_PHASE2)
+endif
+endif
+
+
+# GCP AMI region definition
```
Bajzathd, line 245: AMI means Amazon Machine Images,so there is no such thing as GCP AMI

Edit: I see it was just moved so it is just a nice to have as a fix/improvement

Bajzathd (approved):

pkedvessy (approved):

Makefile
```diff
@@ -214,6 +225,28 @@ Switzerland North:cldrswitzerlandnorth,\
 Germany West Central:cldrgermanywestcentral,\
 Norway East:cldrnorwayeast
 endef
+
+ifndef AZURE_STORAGE_ACCOUNTS
+ifdef IMAGE_COPY_PHASE
```
pkedvessy, line 230: You can simplify L230-L237 with the following:
`if ($(IMAGE_COPY_PHASE),test)
AZURE_STORAGE_ACCOUNTS=$(AZURE_STORAGE_ACCOUNTS_TEST_PHASE)
else
AZURE_STORAGE_ACCOUNTS=$(AZURE_STORAGE_ACCOUNTS_ALL)
endif
`

Its not needed to check IMAGE_COPY_PHASE to be defined as you define it with default value in L14

PR #861, opened 2023-07-26
CB-22606: Reverted Kerberos caching behavior to be backward compatible

TheTinkerDad:


lacikaaa (commented):
- why is this not in the base part instead of final stage?
- why don't you just remove the whole file?

TheTinkerDad:
> * why is this not in the base part instead of final stage?

The file gets installed later than the base part (for Runtime images it gets created by  saltstack/hortonworks/salt/kerberos/kerberos.sls, for FreeIPA it gets created during the IPA server installation)

> * why don't you just remove the whole file?

I'm not familiar with the service's reaction for deleting the whole file. Documentation clearly says "To disable the KCM credential cache, comment out the following lines." so I simply followed documentation.

lacikaaa:
> The file gets installed later than the base part (for Runtime images it gets created by saltstack/hortonworks/salt/kerberos/kerberos.sls, for FreeIPA it gets created during the IPA server installation)

thanks, makes sense in this case

> I'm not familiar with the service's reaction for deleting the whole file. Documentation clearly says "To disable the KCM credential cache, comment out the following lines." so I simply followed documentation.

it shouldn't cause any issues, as only this 2 lines are in the config, and this is just an additional config file under conf.d

Bajzathd (approved):

PR #864, opened 2023-08-01
CB-22635 CIS: Re-use centos7 hardening states for rhel8

Bajzathd:
Reports:

1. RHEL8: https://docs.google.com/spreadsheets/d/1CBUNJ6xlUEBHYgfUYbY59aN5vkRH6RsDSOeyghzMwrg/edit#gid=353245933
266 checks passed, 31 failed and 15 warnings
2. Centos7: https://docs.google.com/spreadsheets/d/1erKV6axNVm1V8CzBCkPrtoiTRI3kpo9JWQ4R7VZMiR4/edit#gid=1466114789
286 checks passed, 29 failed and 11 warnings


scripts/salt-install.sh
```diff
@@ -71,6 +71,8 @@ function update_yum_repos() {
     REPO_FILE=rhel${RHEL_VERSION}_cldr_mirrors.repo
     rm /etc/yum.repos.d/*.repo -f
     curl https://mirror.infra.cloudera.com/repos/rhel/server/8/${RHEL_VERSION}/${REPO_FILE} > /etc/yum.repos.d/${REPO_FILE}
+    # remove whitespaces around '=' as CIS checks do not handle those
```
pkedvessy, line 74: As it is a result of a CIS rule, It would make more sense to have it in the common.sls in a generic way to apply the replace for each repo files.

pkedvessy (approved):

PR #893, opened 2023-09-12
CB-23077 Install dnsutils on non Azure providers

pkedvessy:


scripts/salt-install.sh
```diff
@@ -74,7 +74,9 @@ function update_yum_repos() {
 
     # Workaround on resolving the hostname as for some reason the DNS can't resolve it at provision time
     if [ "${IMAGE_BURNING_TYPE}" == "base" ] ; then
-      yum install -y dnsutils
+      if [[ "${CLOUD_PROVIDER}" != "Azure" ]] ; then
```
Bajzathd, line 77: I think it would make more sense to check if it is already installed, rather than checking for the provider. Or even better, the nobest flag may be enough according to the error message
```
    arm-redhat8:   - cannot install the best candidate for the job
    arm-redhat8: (try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

pkedvessy: I had a try, Its failing with the --nobest flag.

Bajzathd (approved):

PR #919, opened 2023-11-15
CB-23799 xxd changed to od because on rhel8 there is no xxd installed…

doktoric:
…. xxd part of vim-common which contains several vulnerability so we removed that.


TheTinkerDad (approved):
LGTM, but test images are still in progress, hence added the label

PR #926, opened 2023-11-23
CB-23919: Added standalone image copy script for AWS

TheTinkerDad:


scripts/img-copy.sh
```diff
@@ -0,0 +1,69 @@
...
+#    AMIs is enabled for this account. To publicly share the image, you must call 
+#    the DisableImageBlockPublicAccess API."
+# Having public AMIs is a must though, as Cloudbreak needs to access them, so you'll 
+# have to enable this in the account.
+#
+# Note: Regardless of the above error message, the AMI copy should be successful.
+
+if [[ -z $1 ]]; then
+  echo AWS Credentials profile must be provided as the second parameter!
+  exit -1
```
pkedvessy, line 37: As far as I know exit code must be in the 0 - 255 range. Am I wrong? -1 is going to be mapped to 255 so if you really need 255 then plese use 255.

pkedvessy (approved):

PR #939, opened 2023-12-14
CB-23272 In case of RHEL8 and 7.2.18+, use the latest internal repo instead of the 8.8 specific one so we can install Java21

pkedvessy:


saltstack/hortonworks/salt/java/init.sls
```diff
@@ -34,6 +34,24 @@ install_openjdk:
...
+    - pkgs:
+      - java-21-openjdk-headless
+      - java-21-openjdk-devel
+
+delete_rhel89_repo:
+  file.absent:
+    - name: /etc/yum.repos.d/rhel8_cldr_mirrors.repo
+{% endif %}
+
 {# Temporarily removed - will adjust and re-enable once JDK21 gets a RedHat package repo
```
Bajzathd, line 55: I think this block can be removed

Bajzathd (approved):

PR #971, opened 2024-03-18
CB-24163, CB-24164, CB-24165, CB-24166, CB-24175: LUKS volume creation

szabolcs-horvath:
On first boot (started by user-data-helper.sh):
1. Generate passphrase
2. Encrypt passphrase and store ciphertext
3. Create backing file
4. Create loop device on the backing file
5. Create LUKS volume on the loop device
6. Open LUKS volume
7. Create file system inside the LUKS volume
8. Mount LUKS volume

On subsequent boots (done by reopen-luks-volume.service):
1. Recreate loop device on the backing file
2. Decrypt passphrase ciphertext
3. Reopen LUKS volume
4. Remount LUKS volume

Dependencies added:
- cryptsetup
- aws-encryption-sdk-cli
- Python 3.8 to RHEL8 freeipa images (needed for aws-encryption-sdk-cli)

### DIFF
--- a/saltstack/base/salt/luks/bin/create-luks-volume.sh
+++ b/saltstack/base/salt/luks/bin/create-luks-volume.sh
@@ -1,0 +1,156 @@
+#!/bin/bash
+
+set -ex -o pipefail
+
+export LUKS_VOLUME_NAME="cdp-luks"
+export MOUNT_POINT="/mnt/$LUKS_VOLUME_NAME"
+export LUKS_DIR="/etc/$LUKS_VOLUME_NAME"
+export LUKS_BACKING_FILE="$LUKS_DIR/$LUKS_VOLUME_NAME"
+export PASSPHRASE_TMPFS="/mnt/cdp-luks_passphrase_tmpfs"
+export PASSPHRASE_PLAINTEXT="$PASSPHRASE_TMPFS/passphrase"
+export PASSPHRASE_CIPHERTEXT="$LUKS_DIR/passphrase_ciphertext"
+export LUKS_LOG_DIR="/var/log/$LUKS_VOLUME_NAME"
+export LUKS_BACKUP_DIR="$LUKS_DIR/backup"
+export LUKS_MAPPER_DEVICE="/dev/mapper/$LUKS_VOLUME_NAME"
+export ENCRYPTION_KEY_FILE="$LUKS_DIR/passphrase_encryption_key"
+export LUKS_BACKING_FILE_DEFAULT_SIZE="100MiB"
+
+setup_backing_file() {
+  if [[ -e "$LUKS_BACKING_FILE" && $(stat -c "%a" "$LUKS_BACKING_FILE" | tr -d '\n') == 600 ]]; then
+    # Set the backing file to the specified size if it exists
+    truncate -s "$LUKS_BACKING_FILE_DEFAULT_SIZE" "$LUKS_BACKING_FILE"
+  else
+    # Create the backing file if it doesn't exist
+    dd if=/dev/zero of="$LUKS_BACKING_FILE" bs="$LUKS_BACKING_FILE_DEFAULT_SIZE" count=1
+    chmod 600 "$LUKS_BACKING_FILE"
+  fi
+}
+
+setup_tmpfs_for_plaintext_passphrase() {
+  if ! mountpoint "$PASSPHRASE_TMPFS"; then
+    # Create the tmpfs for the plaintext passphrase
+    mount -t tmpfs -o size=1k,mode=700 tmpfs "$PASSPHRASE_TMPFS"
+  fi
+}
+
+generate_passphrase() {
+  if [[ -e "$PASSPHRASE_PLAINTEXT" ]]; then
+    echo "Plaintext passphrase already exists... Exiting LUKS volume creation with failed exit code!"
+    exit 1
+  else
+    aws kms generate-random \
+        --number-of-bytes 64 \
+        --output text \
+        --query Plaintext \
+      | base64 --decode \
+      > "$PASSPHRASE_PLAINTEXT"
+      chmod 600 "$PASSPHRASE_PLAINTEXT"
+  fi
+}
+
+encrypt_passphrase_ciphertext() {
+  INSTANCE_ID="$(TOKEN=$(curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 10") && \
+                                  curl -s -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id)"
+  METADATA_LOG_FILE="$LUKS_LOG_DIR/passphrase_encryption_md-$(date +"%F-%T").json"
+  # Encrypt the plaintext passphrase and store the ciphertext
+  if ! /usr/local/bin/aws-encryption-cli \
+         --encrypt \
+         --input "$PASSPHRASE_PLAINTEXT" \
+         --output "$PASSPHRASE_CIPHERTEXT" \
+         --wrapping-keys provider=aws-kms key="$(cat "$ENCRYPTION_KEY_FILE")" \
+         --metadata-output "$METADATA_LOG_FILE" \
+         --encryption-context INSTANCE_ID="$INSTANCE_ID"; then
+    echo "Failed to encrypt plaintext passphrase... Exiting LUKS volume creation with failed exit code!"
+    exit 2
+  fi
+  chmod 600 "$PASSPHRASE_CIPHERTEXT"
+  chmod 600 "$METADATA_LOG_FILE"
+}
+
+setup_loop_device() {
+  # Create the loop device
+  LOOP_DEVICE=$(losetup --find --show "$LUKS_BACKING_FILE")
+  if [[ $(losetup -j "$LUKS_BACKING_FILE" | cut -d ':' -f1 | tr -d '\n') != "$LOOP_DEVICE" ]]; then
+    echo "Failed to set up the loop device correctly... Exiting LUKS volume creation with failed exit code!"
+    exit 3
+  fi
+}
+
+create_luks_volume() {
+  # Create the LUKS volume
+  if ! echo "YES" | cryptsetup luksFormat "$LOOP_DEVICE" \
+                     --key-file "$PASSPHRASE_PLAINTEXT" \
+                     --type luks2 \
+                     --hash "sha3-512" \
+                     --cipher "aes-xts-plain64" \
+                     --use-random \
+                     --iter-time 2000 \
+                     --pbkdf "pbkdf2" \
+                     --debug; then
+    echo "Failed to create the LUKS volume... Exiting LUKS volume creation with failed exit code!"
+    exit 4
+  fi
+}
+
+open_luks_volume() {
+  # Open the LUKS volume
+  if ! cryptsetup open "$LOOP_DEVICE" "$LUKS_VOLUME_NAME" \
+           --key-file "$PASSPHRASE_PLAINTEXT" \
+           --type luks2 \
+           --debug; then
+    echo "Failed to open the LUKS volume... Exiting LUKS volume creation with failed exit code!"
+    exit 5
+  fi
+}
+
+create_fs_on_luks_volume() {
+  # Create the file system on the LUKS volume
+  if ! mkfs.xfs "$LUKS_MAPPER_DEVICE"; then
+    echo "Failed to create the file system on the LUKS volume... Exiting LUKS volume creation with failed exit code!"
+    exit 6
+  fi
+}
+
+backup_luks_header() {
+  # Create backup of LUKS volume header
+  if ! cryptsetup luksHeaderBackup "$LOOP_DEVICE" \
+                    --header-backup-file "$LUKS_BACKUP_DIR/luks_header_backup-$(date +"%F-%T")" \
+                    --debug; then
+    echo "Failed to create backup of the header of the LUKS volume... Without backup, if the header gets corrupted, the entire volume will become undecryptable!"
+  fi
+}
+
+mount_luks_volume() {
+  if ! mountpoint -q "$MOUNT_POINT"; then
+    # Mount the LUKS volume
+    mount "$LUKS_MAPPER_DEVICE" "$MOUNT_POINT"
+    chmod 700 "$MOUNT_POINT"
+  else
+    echo "Failed to mount the LUKS volume, as the path is already a mount point... Exiting LUKS volume creation with failed exit code!"
+    exit 7
+  fi
+}
+
+main() {
+  if mountpoint "$MOUNT_POINT"; then
+    echo "Volume already mounted at mount point... Exiting LUKS volume creation!"
+    exit
+  fi
+
+  setup_backing_file
+  setup_tmpfs_for_plaintext_passphrase
+  generate_passphrase
+  encrypt_passphrase_ciphertext
+  setup_loop_device
+  create_luks_volume
+  open_luks_volume
+  create_fs_on_luks_volume
+  backup_luks_header
+  mount_luks_volume
+
+  # If everything went fine up to this point, we can enable the reopen service,
+  # the guards in the unit file will stop it from actually executing the reopen script
+  systemctl enable cdp-reopen-luks-volume.service
+}
+
+main
--- a/saltstack/base/salt/luks/bin/reopen-luks-volume.sh
+++ b/saltstack/base/salt/luks/bin/reopen-luks-volume.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+set -ex -o pipefail
+
+export LUKS_VOLUME_NAME="cdp-luks"
+export MOUNT_POINT="/mnt/$LUKS_VOLUME_NAME"
+export LUKS_DIR="/etc/$LUKS_VOLUME_NAME"
+export LUKS_BACKING_FILE="$LUKS_DIR/$LUKS_VOLUME_NAME"
+export PASSPHRASE_TMPFS="/mnt/cdp-luks_passphrase_tmpfs"
+export PASSPHRASE_PLAINTEXT="$PASSPHRASE_TMPFS/passphrase"
+export PASSPHRASE_CIPHERTEXT="$LUKS_DIR/passphrase_ciphertext"
+export LUKS_LOG_DIR="/var/log/$LUKS_VOLUME_NAME"
+export LUKS_MAPPER_DEVICE="/dev/mapper/$LUKS_VOLUME_NAME"
+export ENCRYPTION_KEY_FILE="$LUKS_DIR/passphrase_encryption_key"
+
+recreate_loop_device() {
+  if [[ $(losetup -j "$LUKS_BACKING_FILE" | wc -l | tr -d '\n') == 0 ]]; then
+    # Recreate the loop device
+    LOOP_DEVICE=$(losetup --find --show "$LUKS_BACKING_FILE")
+
+    if [[ $(losetup -j "$LUKS_BACKING_FILE" | cut -d ':' -f1 | tr -d '\n') != "$LOOP_DEVICE" ]]; then
+      echo "Failed to set up loop device correctly... Exiting LUKS volume reopen script with failed exit code!"
+      exit 1
+    fi
+  else
+    LOOP_DEVICE=$(losetup -j "$LUKS_BACKING_FILE" | cut -d ':' -f1 | tr -d '\n')
+  fi
+}
+
+setup_tmpfs_for_plaintext_passphrase() {
+  if ! mountpoint "$PASSPHRASE_TMPFS"; then
+    # Create the tmpfs for the plaintext passphrase
+    mount -t tmpfs -o size=1k,mode=700 tmpfs "$PASSPHRASE_TMPFS"
+  fi
+}
+
+decrypt_passphrase_ciphertext() {
+  if [[ ! -s "$PASSPHRASE_PLAINTEXT" ]]; then
+    INSTANCE_ID="$(TOKEN=$(curl -s -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 10") && \
+                                curl -s -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id)"
+    METADATA_LOG_FILE="$LUKS_LOG_DIR/passphrase_decryption_md-$(date +"%F-%T").json"
+    # Decrypt the ciphertext passphrase
+    if ! /usr/local/bin/aws-encryption-cli \
+           --decrypt \
+           --input "$PASSPHRASE_CIPHERTEXT" \
+           --output "$PASSPHRASE_PLAINTEXT" \
+           --wrapping-keys provider=aws-kms key="$(cat "$ENCRYPTION_KEY_FILE")" \
+           --metadata-output "$METADATA_LOG_FILE" \
+           --encryption-context INSTANCE_ID="$INSTANCE_ID"; then
+      echo "Failed to decrypt the plaintext ciphertext... Exiting LUKS volume reopen script with failed exit code!"
+      exit 2
+    fi
+    chmod 600 "$PASSPHRASE_PLAINTEXT"
+    chmod 600 "$METADATA_LOG_FILE"
+  fi
+}
+
+reopen_luks_volume() {
+  if ! cryptsetup status "$LUKS_VOLUME_NAME"; then
+    # Reopen the LUKS volume
+    if ! cryptsetup open "$LOOP_DEVICE" "$LUKS_VOLUME_NAME" \
+             --key-file "$PASSPHRASE_PLAINTEXT" \
+             --type luks2 \
+             --debug; then
+      echo "Failed to reopen the LUKS volume... Exiting LUKS volume reopen script with failed exit code!"
+      exit 3
+    fi
+  fi
+}
+
+remount_luks_volume() {
+  if ! mountpoint "$MOUNT_POINT"; then
+    # Remount the LUKS volume
+    mount "$LUKS_MAPPER_DEVICE" "$MOUNT_POINT"
+    chmod 700 "$MOUNT_POINT"
+  else
+    echo "Did not mount the LUKS volume, as the path is already a mount point... Exiting LUKS volume reopen script with failed exit code!"
+    exit 4
+  fi
+}
+
+main() {
+  recreate_loop_device
+  setup_tmpfs_for_plaintext_passphrase
+  decrypt_passphrase_ciphertext
+  reopen_luks_volume
+  remount_luks_volume
+}
+
+main
--- a/saltstack/base/salt/luks/etc/systemd/system/cdp-reopen-luks-volume.service
+++ b/saltstack/base/salt/luks/etc/systemd/system/cdp-reopen-luks-volume.service
@@ -1,0 +1,23 @@
+[Unit]
+Description=CDP-LUKS volume reopen script service
+ConditionPathIsMountPoint=!/mnt/cdp-luks
+ConditionFileNotEmpty=/etc/cdp-luks/passphrase_encryption_key
+ConditionFileNotEmpty=/etc/cdp-luks/passphrase_ciphertext
+#TODO Add a Before directive here for services, which will have secrets stored in the LUKS volume
+#Note: Adding a Before directive for a service should only be absolutely necessary,
+#      if otherwise it would start before or in parallel with this service
+#      (for example, if it also starts when network-online.target is reached)
+After=network-online.target
+Wants=network-online.target
+StartLimitIntervalSec=600
+StartLimitBurst=5
+
+[Service]
+Type=oneshot
+ExecStart=/etc/cdp-luks/bin/reopen-luks-volume.sh
+RemainAfterExit=yes
+Restart=on-failure
+TimeoutSec=30
+
+[Install]
+WantedBy=network-online.target
--- a/saltstack/base/salt/luks/init.sls
+++ b/saltstack/base/salt/luks/init.sls
@@ -1,0 +1,71 @@
+/etc/cdp-luks:
+  file.directory:
+    - name: /etc/cdp-luks
+    - user: root
+    - group: root
+    - mode: 700
+
+/etc/cdp-luks/bin:
+  file.directory:
+    - name: /etc/cdp-luks/bin
+    - user: root
+    - group: root
+    - mode: 700
+
+/etc/cdp-luks/backup:
+  file.directory:
+    - name: /etc/cdp-luks/backup
+    - user: root
+    - group: root
+    - mode: 700
+
+/mnt/cdp-luks:
+  file.directory:
+    - name: /mnt/cdp-luks
+    - user: root
+    - group: root
+    - mode: 700
+
+/mnt/cdp-luks_passphrase_tmpfs:
+  file.directory:
+    - name: /mnt/cdp-luks_passphrase_tmpfs
+    - user: root
+    - group: root
+    - mode: 700
+
+/var/log/cdp-luks:
+  file.directory:
+    - name: /var/log/cdp-luks
+    - user: root
+    - group: root
+    - mode: 700
+
+/etc/cdp-luks/bin/create-luks-volume.sh:
+  file.managed:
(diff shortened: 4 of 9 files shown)


saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,98 @@
+/etc/luks:
+  file.directory:
+    - name: /etc/luks
```
lajosrodek, line 3: General: Be sure to update paths here as per the above shell scripts.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,98 @@
...
+    - mode: 600
+    - create: true
+
+passphrase_encryption_key_file:
+  file.managed:
+    - name: /etc/luks/passphrase_encryption_key
+    - user: root
+    - group: root
+    - mode: 600
+    - create: true
```
lajosrodek, line 44: For directories it makes sense, but what is the benefit of creating these zero-byte placeholders here? They will be prepared and overwritten by `create-luks-volume.sh` anyway.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,98 @@
...
+  file.managed:
+    - name: /etc/luks/bin/reopen-luks-volume.sh
+    - source: salt://{{ slspath }}/bin/reopen-luks-volume.sh
+    - user: root
+    - group: root
+    - mode: 700
+
+luks_reopen_service:
+  file.managed:
+    - name: /etc/systemd/system/reopen-luks-volume.service
```
lajosrodek, line 85: Again, please use something like `cdp-reopen-luks-volume.service`.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,98 @@
...
+    - user: root
+    - group: root
+    - makedirs: True
+    - source: salt://{{ slspath }}/etc/systemd/system/reopen-luks-volume.service
```
lajosrodek, line 89: What is the file mode by default? I guess 644 is fine if that is the default one.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,98 @@
...
+
+enable_luks-reopen_service:
+  cmd.run:
+    - name: systemctl enable reopen-luks-volume.service
```
lajosrodek, line 93: This certainly works, by I think there is a separate Salt state (something like `service.enable`?). There are examples in the CB repository.

saltstack/base/salt/prerequisites/packages.sls
```diff
@@ -64,6 +64,7 @@ packages_install:
       - sysstat
       - goaccess
       - httpd-tools
+      - cryptsetup
```
lajosrodek, line 67: I think we only need this for RHEL8 (and later for any subsequent RHEL version meant for AWS GovCloud). So we should like put this inside `{% if pillar['OS'] == 'redhat8' %}` (line 51).

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -72,6 +72,12 @@ reload_sysconf() {
   sysctl -p
 }
```
lajosrodek, line 74: General: Be sure to update LUKS paths here as per the above shell scripts.

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -253,6 +259,11 @@ resize_partitions() {
...
 
 main() {
+  if [[ "$SECRET_ENCRYPTION_ENABLED" == "true" ]]; then
+    echo "$SECRET_ENCRYPTION_KEY_SOURCE" 1> /etc/luks/passphrase_encryption_key
```
lajosrodek, line 263: Nit: Out of curiosity, why specify the file descriptor explicitly? `>` by default operates on stdout / file descriptor 1.

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+if [[ -e /mnt/luks_passphrase_tmpfs/passphrase ]]; then
+  echo "Plaintext passphrase already exists... Exiting LUKS volume creation with failed exit code!"
+  exit 1
+else
+  aws kms generate-random \
+      --number-of-bytes 64 \
+      --output text \
+      --query Plaintext \
+    | base64 --decode \
+    1> /mnt/luks_passphrase_tmpfs/passphrase
```
lajosrodek, line 37: Nit: Out of curiosity, why specify the file descriptor explicitly? `>` by default operates on stdout / file descriptor 1.

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -253,6 +259,11 @@ resize_partitions() {
...
+    create_luks_volume
+    # then move secrets to LUKS volume and symlink them to their original place
```
lajosrodek, line 265: And also extract the EC2 user data secrets and store them either as ciphertext on the root disk, or as plaintext inside the LUKS volume. 😄

saltstack/base/salt/top.sls
```diff
@@ -16,6 +16,7 @@ base:
     - ccmv2
     - custom
     - mount
+    - luks
```
lajosrodek, line 19: Like for `packages.sls`, I wonder if this should be placed in a condition block checking for RHEL8.

But I guess this is not overly critical. `reopen-luks-volume.service` and the `pip` install of `aws-encryption-sdk-cli` are likely compatible with CentOS7 as well, even if never actually used later. 😄

saltstack/freeipa/salt/ipaconsistency/init.sls
```diff
@@ -12,7 +12,7 @@
     - name: /usr/local/lib/python3.8/site-packages/checkipaconsistency/main.py
     - source: salt://{{ slspath }}/scripts/main.py
     - mode: 644
-    - onlyif: ls -la /usr/local/lib/python3.8/site-packages/
+    - onlyif: ls -la /usr/local/lib/python3.8/site-packages/ && ! ls -la /usr/local/lib/python3.6/site-packages/
```
lajosrodek, line 15: Why is this necessary? How did it work before? Was this actually ever executed as I see FreeIPA used to have Python 3.6 only?

lajosrodek (changes requested):
Thanks. A few asks.

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+export LUKS_BACKING_FILE_DEFAULT_SIZE="100MiB"
+
+if mountpoint /mnt/luks; then
+  echo "Volume already mounted at mount point... Exiting LUKS volume creation!"
+  exit
+fi
+
+if [[ -e /etc/luks/luks && $(stat -c "%a" /etc/luks/luks | tr -d '\n') == 600 ]]; then
+  # Set the backing file to the specified size if it exists
+  truncate -s "$LUKS_BACKING_FILE_DEFAULT_SIZE" /etc/luks/luks
```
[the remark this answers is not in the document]
lajosrodek: Oh, I guess this is because `saltstack/base/salt/luks/init.sls` was supposed to pre-create the LUKS image file placeholder? I still think it is better to create the file here.

saltstack/freeipa/salt/ipaconsistency/init.sls
szabolcs-horvath: When there is both Python 3.6 and Python 3.8, the 3.8 salt state failed, because the site-packages directory was only created for 3.6. I added this as a quick workaround to solve this. I am not sure if we need the `checkipaconsistency/main.py` file in both site-packages directories.

saltstack/base/salt/luks/init.sls
szabolcs-horvath: I used these as a way of enforcing 600 permissions.

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+# Create the file system on the LUKS volume
+if ! mkfs.xfs /dev/mapper/luks; then
+  echo "Failed to create the file system on the LUKS volume... Exiting LUKS volume creation with failed exit code!"
+  exit 6
+fi
+
+# Create backup of LUKS volume header
+if ! cryptsetup luksHeaderBackup "$LOOP_DEVICE" \
+                  --header-backup-file "/etc/luks/backup/luks_header_backup-$(date +"%F-%T")" \
+                  --debug; then
```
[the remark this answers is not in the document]
szabolcs-horvath: Yes, this header backup makes it possible to restore the volume, if certain parts of the luks header get damaged. The restoration is done with cryptsetup like this: `cryptsetup luksHeaderRestore --header-backup-file <file> <device>`
A full backup of the volume is still needed though.

saltstack/base/salt/luks/bin/reopen-luks-volume.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -ex -o pipefail
+
+if [[ $(losetup -j /etc/luks/luks | wc -l | tr -d '\n') == 0 ]]; then
+  # Recreate the loop device
```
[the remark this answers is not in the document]
szabolcs-horvath: The loop devices get detached during reboot. This command checks if there is already a loop device associated with the luks backing file. Normally, there shouldn't be any loop devices associated with the backing file after a reboot, but it is worth double checking to make sure we don't set up a duplicate loop device, if for some reason there is already one.

saltstack/base/salt/luks/bin/reopen-luks-volume.sh
```diff
@@ -0,0 +1,47 @@
...
+    exit 3
+  fi
+fi
+
+if ! mountpoint /mnt/luks; then
+  # Remount the LUKS volume
+  mount /dev/mapper/luks /mnt/luks
+  chmod 700 /mnt/luks
+else
+  echo "Did not mount the LUKS volume, as the path is already a mount point..."
```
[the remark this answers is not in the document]
szabolcs-horvath: At this point in the execution the LUKS volume should be fully set up. The earlier commands would have failed with some kind of "device is busy" error if the volume was already mounted at that point. Therefore if for some reason the volume is mounted when the this check runs, there shouldn't be any problem, if we don't try to mount it again. This is unlikely to happen to be fair. Also it might be worth exiting with an error code here just to make sure the extra `chmod` is guaranteed to run. I haven't fully tested if the extra permission setting is always needed, tho.

saltstack/base/salt/luks/init.sls
szabolcs-horvath: The default is defined by the umask of the salt process, which for our case looks to `0022`, so the default should then be `755` then. In that case setting the permissions to something like `644` might make sense here.

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
szabolcs-horvath: No particular reason, could be simply `>` I think.

saltstack/base/salt/luks/bin/reopen-luks-volume.sh
```diff
@@ -0,0 +1,47 @@
...
+
+  if [[ $(losetup -j /etc/luks/luks | cut -d ':' -f1 | tr -d '\n') != "$LOOP_DEVICE" ]]; then
+    echo "Failed to set up loop device correctly... Exiting LUKS volume reopen script with failed exit code!"
+    exit 1
+  fi
+fi
+
+if [[ ! -s /mnt/luks_passphrase_tmpfs/passphrase ]]; then
+  # Decrypt the ciphertext passphrase
+  install -m 600 /dev/null /mnt/luks_passphrase_tmpfs/passphrase
```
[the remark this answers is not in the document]
szabolcs-horvath: You are right, we need to remount the tmpfs here. This could have been a tricky bug which would have placed the plaintext on the disk! Good catch :D

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+
+if [[ -e /mnt/luks_passphrase_tmpfs/passphrase ]]; then
+  echo "Plaintext passphrase already exists... Exiting LUKS volume creation with failed exit code!"
+  exit 1
+else
+  aws kms generate-random \
+      --number-of-bytes 64 \
+      --output text \
+      --query Plaintext \
+    | base64 --decode \
```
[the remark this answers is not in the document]
szabolcs-horvath:
```console
[cloudbreak@ipaserver0 ~]$ cat test.txt | base64 --decode > out.base64
+ base64 --decode
+ cat test.txt
++ printf '\033]0;%s@%s:%s\007' cloudbreak ipaserver0 '~'
[cloudbreak@ipaserver0 ~]$
```
It looks fine.

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+if ! cryptsetup open "$LOOP_DEVICE" luks \
+         --key-file /mnt/luks_passphrase_tmpfs/passphrase \
+         --type luks2 \
+         --debug; then
+  echo "Failed to open the LUKS volume... Exiting LUKS volume creation with failed exit code!"
+  exit 5
+fi
+
+# Create the file system on the LUKS volume
+if ! mkfs.xfs /dev/mapper/luks; then
```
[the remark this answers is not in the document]
szabolcs-horvath: `/dev/mapper/luks` is the block device created by cryptsetup. It's name is determined by the argument given to `cryptsetup open`, so I will extract this to an env var.

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+    1> /mnt/luks_passphrase_tmpfs/passphrase
+fi
+
+# Encrypt the plaintext passphrase and store the ciphertext in persistent storage
+if ! /usr/local/bin/aws-encryption-cli \
+       --encrypt \
+       --input /mnt/luks_passphrase_tmpfs/passphrase \
+       --output /etc/luks/passphrase_ciphertext \
+       --wrapping-keys provider=aws-kms key="$SECRET_ENCRYPTION_KEY_SOURCE" \
+       --metadata-output "/var/log/luks/passphrase_encryption_md-$(date +"%F-%T").json" \
```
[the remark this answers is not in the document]
szabolcs-horvath: > Metadata Contents[](https://aws-encryption-sdk-cli.readthedocs.io/en/latest/#metadata-contents)
> The metadata JSON contains the following fields:
> - "mode" : "encrypt"/"decrypt"/"decrypt-unsigned"
> - "input" : Full path to input file (or "<stdin>" if stdin)
> - "output" : Full path to output file (or "<stdout>" if stdout)
> - "header" : JSON representation of [message header data](http://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/message-format.html#header-structure)
> - "header_auth" : JSON representation of [message header authentication data](http://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/message-format.html#header-authentication) (only on decrypt)

I don't think anything here is sensitive, but I'm not 100% sure.

saltstack/base/salt/luks/bin/reopen-luks-volume.sh
```diff
@@ -0,0 +1,47 @@
...
+  if ! /usr/local/bin/aws-encryption-cli \
+         --decrypt \
+         --input /etc/luks/passphrase_ciphertext \
+         --output /mnt/luks_passphrase_tmpfs/passphrase \
+         --wrapping-keys provider=aws-kms key="$(cat /etc/luks/passphrase_encryption_key)" \
+         --metadata-output "/var/log/luks/passphrase_decryption_md-$(date +"%F-%T").json" \
```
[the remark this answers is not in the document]
szabolcs-horvath: See previous comment about the metadata JSON.

saltstack/base/salt/luks/etc/systemd/system/reopen-luks-volume.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=LUKS volume reopen script service
+ConditionPathIsMountPoint=!/mnt/luks
+ConditionFileNotEmpty=/etc/luks/passphrase_encryption_key
+ConditionFileNotEmpty=/etc/luks/passphrase_ciphertext
+After=cloud-init.target
```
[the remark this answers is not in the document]
szabolcs-horvath:
```sls
start-cloudera-scm-server:
  service.running:
    - name: cloudera-scm-server
```
I guess a salt highstate starts it after a reboot.

saltstack/base/salt/luks/bin/create-luks-volume.sh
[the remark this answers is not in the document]
szabolcs-horvath:
```json
{
  "header": {
    "algorithm": "AES_256_GCM_HKDF_SHA512_COMMIT_KEY_ECDSA_P384",
    "commitment_key": "VBM/Pw4HRf6jOI15qQEYUkCU0hRLCNheDh7m2OQ/KsU=",
    "content_type": 2,
    "encrypted_data_keys": [
      {
        "encrypted_data_key": "AQIBAHjNY1+vWYKur3ANO12lNTnqtta8fuEMpqkruayWZfi1BgFvdts9YF+1x0fm4ISB3uWRAAAAfjB8BgkqhkiG9w0BBwagbzBtAgEAMGgGCSqGSIb3DQEHATAeBglghkgBZQMEAS4wEQQMUieGo91dbJ0h4M2bAgEQgDtdHaYG8Qbo8KgMe7xwQkgR9zkK4Vd9DCDtPSystBFZJlrMFsenRyPj2n8nPxcGbgvrTUiz66BM/38ncw==",
        "key_provider": {
          "key_info": "YXJuOmF3cy11cy1nb3Y6a21zOnVzLWdvdi13ZXN0LTE6MjAyNTUwMTY1MDcyOmtleS9mOWRlYTg4Yy1lMWM3LTQ3ZGUtOTZiMi03OWIxZTVlMjc1ZDU=",
          "provider_id": "YXdzLWttcw=="
        }
      }
    ],
    "encryption_context": {
      "INSTANCE_ID": "i-0f7902a80e0658a05",
      "aws-crypto-public-key": "Ah/B1SIX+0MrsarccEch7Zil8DJxQmFk953LYsCLgzijlUubQFYvxr90Zn3vhI/ULQ=="
    },
    "frame_length": 4096,
    "header_iv_length": null,
    "message_id": "RIKnSrpuUq74i+T72umNYIeyrk6H6R24FwXB3SXJ+HM=",
    "type": null,
    "version": "2.0"
  },
  "input": "/mnt/cdp-luks_passphrase_tmpfs/passphrase",
  "mode": "encrypt",
  "output": "/etc/cdp-luks/passphrase_ciphertext"
}
```
This is an example metadata JSON. I would need to check if any of this is sensitive, but on first look only the `encrypted_data_key` seems like something that could be sensitive.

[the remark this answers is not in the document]
szabolcs-horvath: Also, seems like the `aws-encryption-sdk-cli` creates the file with `640` permissions by default. Is that restrictive enough, or would it need to be `600` for example?

saltstack/base/salt/luks/bin/create-luks-volume.sh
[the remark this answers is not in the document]
lajosrodek: I was first puzzled by the `++ printf` stuff. I suppose it was the prompt message perhaps? :smile:

saltstack/base/salt/luks/bin/create-luks-volume.sh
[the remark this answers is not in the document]
lajosrodek: I would definitely go with `600` for these ones, for consistency. `encrypted_data_key` can be decrypted with the KMS key, and every cluster user has access to the latter (via the instance profile role Principal in the KMS key policy. Although that is only the data key itself; the passphrase plaintext or ciphertext are not accessible to regular users (only to `root`), so having the data key plaintext would not gain anything. :smile: 

That aside, `encrypted_data_key` is a ciphertext, so there is no harm revealing it in the metadata JSON (if restricted to `root` only), because the KMS key is kept outside the cluster / disk.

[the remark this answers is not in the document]
lajosrodek: `aws-crypto-public-key` seems to be only used for the AAD of the digital signature, so it is not sensitive either.

saltstack/base/salt/luks/bin/create-luks-volume.sh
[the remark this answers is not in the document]
szabolcs-horvath: Same, yeah it's the prompt 😄

saltstack/base/salt/luks/init.sls
szabolcs-horvath: But, yeah it probably makes more sense to create these in the script just before they are needed.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+/etc/cdp-luks/backup:
+  file.directory:
+    - name: /etc/cdp-luks/backup
+    - user: root
+    - group: root
+    - mode: 700
+
+luks_mount_point:
+  file.directory:
+    - name: /mnt/cdp-luks
```
lajosrodek, line 24: Perhaps change the state name to `/mnt/cdp-luks`.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+    - user: root
+    - group: root
+    - mode: 700
+
+luks_passphrase_tmpfs_mount_point:
+  file.directory:
+    - name: /mnt/cdp-luks_passphrase_tmpfs
```
lajosrodek, line 31: Perhaps change the state name to `/mnt/cdp-luks_passphrase_tmpfs`.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+    - user: root
+    - group: root
+    - mode: 700
+
+aws_encryption_sdk_logs:
+  file.directory:
+    - name: /var/log/cdp-luks
```
lajosrodek, line 38: Perhaps change the state name to `/var/log/cdp-luks`.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+    - user: root
+    - group: root
+    - mode: 700
+
+luks_create_script:
+  file.managed:
+    - name: /etc/cdp-luks/bin/create-luks-volume.sh
```
lajosrodek, line 45: Perhaps change the state name to `/etc/cdp-luks/bin/create-luks-volume.sh`.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+    - source: salt://{{ slspath }}/bin/create-luks-volume.sh
+    - user: root
+    - group: root
+    - mode: 700
+
+luks_reopen_script:
+  file.managed:
+    - name: /etc/cdp-luks/bin/reopen-luks-volume.sh
```
lajosrodek, line 53: Perhaps change the state name to `/etc/cdp-luks/bin/reopen-luks-volume.sh`.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+    - source: salt://{{ slspath }}/bin/reopen-luks-volume.sh
+    - user: root
+    - group: root
+    - mode: 700
+
+luks_reopen_service:
+  file.managed:
+    - name: /etc/systemd/system/cdp-reopen-luks-volume.service
```
lajosrodek, line 61: Perhaps change the state name to `/etc/systemd/system/cdp-reopen-luks-volume.service`.

lajosrodek (changes requested):
Thanks. A few more tiny asks.

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,152 @@
...
+  METADATA_LOG_FILE="$LUKS_LOG_DIR/passphrase_encryption_md-$(date +"%F-%T").json"
+  install -m 600 /dev/null "$METADATA_LOG_FILE"
+  install -m 600 /dev/null "$PASSPHRASE_CIPHERTEXT"
+  # Encrypt the plaintext passphrase and store the ciphertext
+  if ! /usr/local/bin/aws-encryption-cli \
+         --encrypt \
+         --input "$PASSPHRASE_PLAINTEXT" \
+         --output "$PASSPHRASE_CIPHERTEXT" \
+         --wrapping-keys provider=aws-kms key="$(cat "$ENCRYPTION_KEY_FILE")" \
+         --metadata-output "$METADATA_LOG_FILE" \
```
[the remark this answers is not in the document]
szabolcs-horvath: It seems it changes `600` to `640`, so I added `chmod`s after it instead.

saltstack/base/salt/luks/bin/reopen-luks-volume.sh
```diff
@@ -0,0 +1,88 @@
...
+    METADATA_LOG_FILE="$LUKS_LOG_DIR/passphrase_decryption_md-$(date +"%F-%T").json"
+    install -m 600 /dev/null "$METADATA_LOG_FILE"
+    install -m 600 /dev/null "$PASSPHRASE_PLAINTEXT"
+    # Decrypt the ciphertext passphrase
+    if ! /usr/local/bin/aws-encryption-cli \
+           --decrypt \
+           --input "$PASSPHRASE_CIPHERTEXT" \
+           --output "$PASSPHRASE_PLAINTEXT" \
+           --wrapping-keys provider=aws-kms key="$(cat "$ENCRYPTION_KEY_FILE")" \
+           --metadata-output "$METADATA_LOG_FILE" \
```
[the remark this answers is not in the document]
szabolcs-horvath: It seems it changes `600` to `640`, so I added `chmod`s after it instead.

saltstack/base/salt/luks/init.sls
```diff
@@ -0,0 +1,75 @@
...
+    - user: root
+    - group: root
+    - mode: 644
+    - makedirs: True
+    - source: salt://{{ slspath }}/etc/systemd/system/cdp-reopen-luks-volume.service
+
+enable_luks_reopen_service:
+  service.enabled:
+    - name: cdp-reopen-luks-volume.service
```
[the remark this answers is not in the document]
szabolcs-horvath: Moved it to the create script, as it is probably safer 😄

saltstack/base/salt/luks/etc/systemd/system/reopen-luks-volume.service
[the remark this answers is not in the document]
szabolcs-horvath: I changed the reopen service to autostart after `network-online.target` is reached as it is indeed the point where the network can actually considered be "up" (see: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/), and we should be able to decrypt the passphrase ciphertext. 

This would mean it would start in parallel with the CM service for example, which is not what we want, so I added a TODO and a note to the unit file, to add a `Before` directive for services like the CM service, to ensure the LUKS volume is mounted before these services would start and try to use secrets stored on the encrypted volume.
I think adding these lines to the unit file can be done when we are actually going through the secrets and creating the symlinks for them.

I assume no service is trying to start before the network is "up", as in that case I do not know what we should do 👀

Bajzathd:
@szabolcs-horvath are there any test images that were created and validated successfully?

szabolcs-horvath:
@Bajzathd I have validated previous versions, but I am currently burning new test images. I will post them here once I validated them.

szabolcs-horvath:
> @Bajzathd I have validated previous versions, but I am currently burning new test images. I will post them here once I validated them.

These are the images:
- freeipa: https://github.infra.cloudera.com/raw/gist/shorvath/183fca88a54adf2e3de28fe77df2f6aa/raw/4a01c9c1edb95555888c4abf9e298faeb440ea84/luks_volume_poc_freeipa_image_catalog.json
- prewarm: https://github.infra.cloudera.com/raw/gist/shorvath/386551cc4daf5211e0c4facc3e79f51d/raw/2f07ad0c1d45bd6ccf681e1c3b33babd4e64ce4b/luks_volume_poc_cb_image_catalog.json

I will probably only get to validating them on Monday, but I'm sending them in case you want to look at them until then.

saltstack/base/salt/luks/etc/systemd/system/reopen-luks-volume.service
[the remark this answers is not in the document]
lajosrodek: Great. I agree we should somehow make sure that `cdp-reopen-luks-volume` starts and completes before `cloudera-scm-server`.

Since we are currently focusing on getting FreeIPA up & running, I suggest raising a task for this under the epic and have it handled later when we get to dealing with DL & DH. Could you please create a Jira ticket for this? :smile:

saltstack/base/salt/luks/etc/systemd/system/cdp-reopen-luks-volume.service
```diff
@@ -0,0 +1,23 @@
...
+Description=CDP-LUKS volume reopen script service
+ConditionPathIsMountPoint=!/mnt/cdp-luks
+ConditionFileNotEmpty=/etc/cdp-luks/passphrase_encryption_key
+ConditionFileNotEmpty=/etc/cdp-luks/passphrase_ciphertext
+#TODO Add a Before directive here for services, which will have secrets stored in the LUKS volume
+#Note: Adding a Before directive for a service should only be absolutely necessary,
+#      if otherwise it would start before or in parallel with this service
+#      (for example, if it also starts when network-online.target is reached)
+After=network-online.target
+Wants=network-online.target
```
lajosrodek, line 11: Looks good. We can fix this later once the work on FreeIPA is done.

lajosrodek (approved):
Thanks, LGTM

saltstack/base/salt/luks/bin/create-luks-volume.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# Create the LUKS volume
+if ! echo "YES" | cryptsetup luksFormat "$LOOP_DEVICE" \
+                   --key-file /mnt/luks_passphrase_tmpfs/passphrase \
+                   --type luks2 \
+                   --hash "sha3-512" \
+                   --cipher "aes-xts-plain64" \
+                   --use-random \
+                   --iter-time 2000 \
+                   --pbkdf "pbkdf2" \
```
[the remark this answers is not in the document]
lajosrodek: [CB-25446](https://jira.cloudera.com/browse/CB-25446) has been raised for this.

saltstack/base/salt/luks/etc/systemd/system/cdp-reopen-luks-volume.service
lajosrodek: [CB-25449](https://jira.cloudera.com/browse/CB-25449) has been raised for this.

saltstack/base/salt/top.sls
Bajzathd: What about cloud providers? Isn't this AWS_GOV only?

saltstack/base/salt/prerequisites/packages.sls
```diff
@@ -52,6 +52,7 @@ packages_install:
     {% endif %}
     {% if pillar['OS'] == 'redhat8' %}
       - sos
+      - cryptsetup
```
Bajzathd, line 55: Is this needed for all cloud providers?

szabolcs-horvath:
Manually validated with the following images:
- freeipa: https://github.infra.cloudera.com/cloudbreak/cb-image-catalog/commit/f9c7f2743baa567fc5fd3be758d9efba6de10a3c
- prewarm: https://github.infra.cloudera.com/cloudbreak/cb-image-catalog/commit/9bd216cedb6b611366ed3a06af509d1049414d08

These images don't contain the small fix in the last force push (https://github.com/hortonworks/cloudbreak-images/compare/67a06e6c7cac30723532e74a2fb83edc5a560736..0a36f78d61f5a33662f459d90e3ab3352ff22f1b), so I applied this change manually while testing the images.

Bajzathd (approved):

TheTinkerDad (approved):

PR #982, opened 2024-04-19
CB-24126: Retrieve userdata secrets and delete secret

szabolcs-horvath:


saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
 
 set -x
-set
+
+if [[ "$SECRET_ENCRYPTION_ENABLED" == "true" ]]; then
+  set | /usr/bin/redact-secrets.sh
+  source /usr/bin/retrieve-userdata-secrets.sh $> /var/log/retrieve-userdata-secrets.log
```
lajosrodek, line 7: Nit: For consistency, perhaps add `cdp-` prefix to `retrieve-userdata-secrets.sh` to make it in sync with `cdp-luks`. Same goes for `retrieve-userdata-secrets.log`.

saltstack/base/salt/userdata-secrets/bin/retrieve-userdata-secrets.sh
```diff
@@ -0,0 +1,68 @@
...
+    # aws cli with standard retry mode uses an exponential backoff with a base factor of 2 and a maximum backoff time of 20 seconds
+    aws_retry_mode=standard
+    # AWS_MAX_ATTEMPTS=15 would mean a maximum wait time of approximately 4 minutes per aws cli command
+    aws_max_attempts=15
+
+    echo "Retrieving userdata secrets..."
+    AWS_RETRY_MODE="$aws_retry_mode" \
+    AWS_MAX_ATTEMPTS="$aws_max_attempts" \
+    retry_until_non_placeholder_value \
+    aws secretsmanager get-secret-value \
```
lajosrodek, line 48: While `retry_until_non_placeholder_value()` inherits the changes of `AWS_RETRY_MODE` and `AWS_MAX_ATTEMPTS`, they are not exposed to the child process started by `SECRET_STRING="$("$@")"`, unless these environment variables are marked for `export`.

I suggest either

- Turning these variable assignments into explicit `export` commands executed on their own (i.e. dropping the backslash at the end of lines 45 and 46); or
- Reording the lines, making `retry_until_non_placeholder_value` the first one so that the variable assignments become part of its arguments.

saltstack/base/salt/userdata-secrets/bin/retrieve-userdata-secrets.sh
```diff
@@ -0,0 +1,68 @@
...
+  done
+  echo "Userdata secret's value was retrieved after waiting $total_wait_time seconds."
+}
+
+# Intermittent issues like throttling of AWS APIs are handled by the built in retry mechanism of the aws cli.
+# Waiting for CB to update the secret's value to the actual secrets is handled by the function above.
+main() {
+  if [[ "$CLOUD_PLATFORM" == "AWS" ]]; then
+    # See: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html
+    # aws cli with standard retry mode uses an exponential backoff with a base factor of 2 and a maximum backoff time of 20 seconds
```
lajosrodek, line 39: Great stuff! Just thinking out loud, the standard retry strategy does not account for access denied errors (presumably 403 or 400, however this is mapped in the CLI response).

This will hopefully not be an issue; the secret resource has initially full access (no resource policy), they the resource policy is set (permitting access to a specific EC2 instance), followed by setting the real secret value. In practice, no access denied shall be encountered when executed from the target EC2 instance, so let us ignore this until if surfaces (if ever). :smile:

saltstack/base/salt/userdata-secrets/bin/redact-secrets.sh
```diff
@@ -0,0 +1,24 @@
...
+while read -r secret; do
+    SECRETS+=("$secret")
+done < $SECRETS_FILE
+
+while read -r line; do
+    matched=false
+    for secret in "${SECRETS[@]}"; do
+        if [[ "$line" =~ ($secret)=.* ]]; then
+            matched=true
+            echo "${BASH_REMATCH[1]}=*SECRET REDACTED*"
```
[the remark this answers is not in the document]
szabolcs-horvath: Right, as I just moved the userdata secret extraction after the call to `set`, this isn't needed anymore, as the actual secret values wont be present yet in the environment variables when set is called. Previously the `set` call would expose the secrets as they were extracted before the `user-data-helper.sh`.

saltstack/base/salt/userdata-secrets/bin/retrieve-userdata-secrets.sh
szabolcs-horvath: Should hopefully be fine, since we atomically update the resource policy and the secret value at once.

saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -3,6 +3,10 @@
 set -x
 set
 
+if [[ "$SECRET_ENCRYPTION_ENABLED" == "true" ]]; then
+  source /usr/bin/cdp-retrieve-userdata-secrets.sh $> /var/log/cdp-retrieve-userdata-secrets.log
```
lajosrodek, line 7: `$>` -> `&>`

lajosrodek (changes requested):
Thanks, only one typo.

lajosrodek (approved):
Thanks, LGTM

lajosrodek (approved):
Still LGTM :smile:

TheTinkerDad (approved):

PR #998, opened 2024-06-26
CB-26201: Add a way to build CentOS7 Runtime/FreeIPA images from a base image

TheTinkerDad:
Test build - AWS, Runtime: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/4758/
This is based on the foundation image burned with https://github.com/hortonworks/cloudbreak-images/pull/999


TheTinkerDad:
As explained on the ticket, this won't be done after all...

PR #1006, opened 2024-07-16
CB-24115 CB-24117 CB-24118 CB-24119 CB-24120 CB-24121 CB-26031 Popula…

lajosrodek:
…te secret storage

* Fix the permission of the LUKS mount point (700 -> 755).
* Add new script `/etc/cdp-luks/bin/populate-luks-volume.sh` for populating the LUKS volume with secrets.
  * Invoked from `user-data-helper.sh` right after the LUKS volume creation.
* Secrets covered here:
  * Salt pillars: Gateway nodes only.
  * CCMv2 configs: FreeIPA nodes only for v2.5 Jumpgate; all the nodes otherwise if CCMv2 is enabled.
  * HTTP Proxy configs: All the nodes if HTTP Proxy is enabled.
  * nginx host certificate private key: Gateway nodes only.
  * `ipahealthagent` certificate private key: FreeIPA nodes only.
  * Salt Bootstrap configs: All the nodes.
  * Salt API certificate private key: Gateway nodes only.
  * Salt Master certificate private keys: Gateway nodes only.
  * Salt Minion certificate private key: All the nodes.
  * FreeIPA services (GSS API, GSS Proxy, custodia, dnssec, NSS DB, Kerberos KDC, Kerberos keytabs): `/etc/ipa/nssdb` for all the nodes; the rest for FreeIPA nodes only.
  * SSSD (SSSD cache, AD trust keytabs, KCM cache): All the nodes.
* Secrets covered later:
  * KDC default keytab (`/etc/krb5.keytab`): See [CB-26477](https://jira.cloudera.com/browse/CB-26477).
  * Kerberos TGT cache (ccache): See [CB-26478](https://jira.cloudera.com/browse/CB-26478).
  * `dirsrv` / `slapd` DB (if Kerberos principal password hashes need to be protected): See [CB-26479](https://jira.cloudera.com/browse/CB-26479).

### DIFF
--- a/saltstack/base/salt/luks/bin/populate-luks-volume.sh
+++ b/saltstack/base/salt/luks/bin/populate-luks-volume.sh
@@ -1,0 +1,168 @@
+#!/bin/bash
+
+set -ex -o pipefail
+
+export LUKS_VOLUME_NAME="cdp-luks"
+export MOUNT_POINT="/mnt/$LUKS_VOLUME_NAME"
+
+validate_absolute_path() {
+  local location="$1"
+  if [[ ! "$location" =~ ^/.* ]]
+  then
+    echo "Location \"$location\" must be an absolute path!"
+    exit 1
+  fi
+}
+
+validate_location_exists() {
+  local location="$1"
+  if [ ! -d "$location" ]
+  then
+    echo "Location \"$location\" must be an existing directory!"
+    exit 2
+  fi
+}
+
+create_location_if_necessary() {
+  local location="$1"
+  local owner=$2
+  local group=$3
+  local mode=$4
+  if [ ! -d "$location" ]
+  then
+    echo "Location \"$location\" does not exist; creating it with owner \"$owner\", group \"$group\", mode \"$mode\"."
+    install -o $owner -g $group -m $mode -d "$location"
+  fi
+}
+
+create_location_parent_tree_if_necessary() {
+  local source_location_parent
+  source_location_parent=$(dirname "$1")
+  if [ "$source_location_parent" != "/" ]
+  then
+    echo "Creating source location parent tree \"$source_location_parent\" if necessary."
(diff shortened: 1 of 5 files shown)


saltstack/base/salt/luks/bin/populate-luks-volume.sh
```diff
@@ -0,0 +1,168 @@
...
+
+log_processing_skipped() {
+  echo "No need to process $1 locations."
+}
+
+process_global_locations() {
+  log_processing_needed "global"
+  process_location_with_create "/etc/salt-bootstrap" root root 700
+  create_location_if_necessary "/etc/salt/pki" root root 744
+  process_location_with_create "/etc/salt/pki/minion" root root 700
```
szabolcs-horvath, line 91: The explicit creation of `/etc/salt/pki` here is needed because it needs to have looser permissions than `/etc/salt/pki/minion`, if I understand correctly?

szabolcs-horvath (approved):
LGTM, only one question

saltstack/base/salt/luks/bin/populate-luks-volume.sh
lajosrodek: That is correct. The permissions seen here are the exact ones currently found in our clusters (checked mostly on a green FreeIPA cluster, or sometimes on a green DL). I felt it more consistent to mirror the current behavior in these LUKS locations. These locations do not exist in the images, they only manifest upon the startup of the respective Salt processes. (I could not find any direct reference to these locations in CB repos, so I presume they are directly created by Salt Minion etc.)

On the other hand, I did not investigate further what the reason is for such discrepancies (like `/etc/salt/pki` vs `/etc/salt/pki/minion`); possibly these are created by separate processes having a different umask, or the same process explicitly using different access mode settings.

TheTinkerDad (approved):

PR #1012, opened 2024-07-28
CB-25319: Integrate/automate STIG compliance checks

sgabor001:


saltstack/final/salt/top.sls
```diff
@@ -11,6 +11,7 @@ final:
 {% if pillar['subtype'] != 'Docker' %}
     - cis-controls
 {% endif %}
+    - openscap
```
Bajzathd, line 14: I don't think that we need this on Docker

saltstack/final/salt/openscap/init.sls
```diff
@@ -0,0 +1,19 @@
+openscap_install:
```
Bajzathd, line 1: A few general comments for this file:

1. I think this should only run when stig is enabled
2. The installed packages should be cleaned up, similarly to ansible in cis-controls
3. I guess the results should be under /tmp/cis too, so that we can see it on Jenkins side

saltstack/final/salt/cis-controls/scripts/cis_control.sh
```diff
@@ -53,14 +53,16 @@ git clone https://github.com/AutomateCompliance/AnsibleCompliancePlaybooks.git $
 ssh-keygen -A
 
 #Apply the SSG Ansible playbook
+mkdir /tmp/cis
+chmod 777 /tmp/cis
```
Bajzathd, line 57: Do we really need 777? Also I guess it could be applied recursively one time at the end of image burning

saltstack/final/salt/top.sls
sgabor001: yes it wast planned

saltstack/final/salt/openscap/init.sls
sgabor001: In the top.sls the stig_enebled var is checked.
I do not think that the openscap install log have to be collected for every image. If install issue occurs during burning time it can be investigated based on the jenkins log.
Maybe the openscap can be useful for newer checks in the future for example during engesc...

saltstack/final/salt/cis-controls/scripts/cis_control.sh
sgabor001: I guess for the check itself not necessary but for the log collection yes. I do not think that it is a problem during burning time.

Bajzathd (approved):

PR #1013, opened 2024-07-29
CB-26336: Configure pip logging on image

sgabor001:


TheTinkerDad (approved):
LGTM, also I suggest using separate log files for each pip installation

PR #1025, opened 2024-08-19
CB-26928: Added security patches for Azure

TheTinkerDad:


kkubina (approved):

TheTinkerDad:
This needs to be extended to other cloud providers and also tested, including FreeIPA images.

PR #1032, opened 2024-08-26
CB-24168: Symlink CM secret locations to secret storage

szabolcs-horvath:


szabolcs-horvath:
Validated with this image: https://github.infra.cloudera.com/cloudbreak/cb-image-catalog/commit/693fc0ce52ed1f8a22786ee42eca1a1f6d7467ef

saltstack/base/salt/luks/bin/populate-luks-volume.sh
```diff
@@ -157,12 +157,23 @@ process_freeipa_locations() {
   fi
 }
 
+process_cm_locations() {
+  if [[ "$IS_FREEIPA" != "true" ]]
+  then
+    log_processing_needed "Cloudera Manager"
+    /opt/cloudera/cm/bin/change_secrets_storage.sh "$MOUNT_POINT"
```
lajosrodek, line 164: I know CM folks added this [OPSAPS-69120](https://jira.cloudera.com/browse/OPSAPS-69120), but it does not currently bring any benefits over invoking `process_location` for every location involved. Looks like that script places everything into the subtree `$SECRET_PATH/cm` instead of `$SECRET_PATH`, which is inconsistent with the directory structure we use here for the other locations.

I thus wonder if we would be better off skip using the CM script and relying on `process_location` instead? (The parent paths `/etc` and `/var/lib` always exist, so no need to use `create_location_if_necessary` or `process_location_with_create`.)

saltstack/base/salt/luks/bin/populate-luks-volume.sh
```diff
@@ -157,12 +157,23 @@ process_freeipa_locations() {
   fi
 }
 
+process_cm_locations() {
+  if [[ "$IS_FREEIPA" != "true" ]]
```
lajosrodek, line 161: This certainly works. On the other hand, I think `/etc/cloudera-scm-server` can only be found on `"$IS_GATEWAY" == "true"` nodes. (Actually, on exactly one gateway node I think, the primary one, because CM Server is not deployed in a HA fashion yet.) The `cloudera-scm-agent` locations are indeed for every node, though.

lajosrodek (changes requested):

saltstack/base/salt/luks/bin/populate-luks-volume.sh
szabolcs-horvath: I checked, and on the `idbroker` instance of the DL for example, the `etc/cloudera-scm-server` directory does exist, and contains the following:
```console
[root@shorvath-aws-dl-idbroker0 cloudera-scm-server]# ls -la
total 40
drwxr-xr-x.   2 cloudera-scm cloudera-scm   98 Aug 27 12:43 .
drwxr-xr-x. 121 root         root         8192 Sep  3 10:30 ..
-rw-r--r--.   1 root         root         8197 Aug 22 00:59 cmSubCaCert.profile
-rw-------.   1 cloudera-scm cloudera-scm  757 Aug 22 00:59 db.properties
-rw-r--r--.   1 root         root         4495 Aug 22 00:59 log4j.properties
-rw-r--r--.   1 root         root           80 Aug 27 12:43 repo_url.txt
```
Of these, the `db.properties` file could contain the CM DB password for example. Although it seems like this file only contains actual values on the gateway instance. Still, it might make sense to protect these files.

szabolcs-horvath: This `28K` in total btw.

lajosrodek: Oh, I see. Then let us leave it as is. :smile:

szabolcs-horvath:
Validated with this image: https://github.infra.cloudera.com/cloudbreak/cb-image-catalog/commit/5e8b5302f50fd0e63a75fd06c2cd9a508707032d

Bajzathd (approved):

PR #1035, opened 2024-08-26
CB-26917 Fix revert in 0a6f900

Bajzathd:


TheTinkerDad (approved):

Bajzathd:
Revert was reverted in https://github.com/hortonworks/cloudbreak-images/pull/1036 so this is no longer needed

PR #1036, opened 2024-08-27
CB-26992 Use latest cdp-infra-tools and its gpg key

Bajzathd:
Azure x86: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/5284/
AWS arm64: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/5285/


saltstack/base/salt/telemetry/init.sls
```diff
@@ -1,2 +1,33 @@
...
+    - name: yum --disablerepo="*" --enablerepo=cdp-infra-tools list available
+
+install_cdp_infra_tools_packages:
+  pkg.installed:
+    - pkgs:
+{% if salt['environ.get']('INCLUDE_CDP_TELEMETRY') == "Yes" %}
+      - cdp-telemetry
+{% endif %}
+{% if salt['environ.get']('INCLUDE_FLUENT') == "Yes" %}
+      - redhat-lsb-core # this will install redhat-lsb-core is required for fluent
```
TheTinkerDad, line 24: This line is redundant, redhat-lsb-core is always installed much earlier as it is needed by other stuff too.

TheTinkerDad (changes requested):

saltstack/base/salt/telemetry/init.sls
TheTinkerDad: Nevermind, actually. Seems like fluent always needs this while the other components only need it on RHEL 8.

TheTinkerDad: https://github.com/hortonworks/cloudbreak-images/blob/86cca4e9571714bc8a0306ff6c1b71671d7e6a0c/scripts/salt-install.sh#L44

Bajzathd: the earlier install is for rhel8 only

TheTinkerDad (approved):

PR #1039, opened 2024-08-29
CB-26510: Workaround for DNS resolution of KMS endpoint for secret encryption passphrase decryption

szabolcs-horvath:
On FreeIPA nodes, DNS resolution is done by the `named-pkcs11` service. This service is not up by the time we are trying to decrypt the passphrase, to remount the LUKS volume by calling AWS KMS, because the files it needs are actually on the encrypted LUKS volume. This creates a chicken and egg problem.

The workaround solution is to use a different DNS server for resolving the KMS endpoint needed for the decryption process:
1. We try the AmazonProvidedDNS server first. If that isn't able to resolve the endpoint to an address, then we try the IP addresses of the DNS servers associated with the VPC the instance is in.
2. After we resolved the endpoint to an address, we add a temporary entry to the `/etc/hosts` file containing the resolved IP and the endpoint. At this point we can decrypt the passphrase.
3. Once the decryption is done, or when it fails, we restore the `/etc/hosts` file to its previous state, using a backup file created when we added the entry to the file.

This is only an issue on FreeIPA nodes, as DL and DH nodes use the FreeIPA in their environment as DNS server.

### DIFF
--- a/saltstack/base/salt/luks/bin/reopen-luks-volume.sh
+++ b/saltstack/base/salt/luks/bin/reopen-luks-volume.sh
@@ -12,7 +12,11 @@ export PASSPHRASE_CIPHERTEXT="$LUKS_DIR/passphrase_ciphertext"
 export LUKS_LOG_DIR="/var/log/$LUKS_VOLUME_NAME"
 export LUKS_MAPPER_DEVICE="/dev/mapper/$LUKS_VOLUME_NAME"
 export ENCRYPTION_KEY_FILE="$LUKS_DIR/passphrase_encryption_key"
-
+{% if pillar['CUSTOM_IMAGE_TYPE'] == 'freeipa' %}
+export IS_FREEIPA=true
+{% else %}
+export IS_FREEIPA=false
+{% endif %}
 export AWS_USE_FIPS_ENDPOINT=true
 
 recreate_loop_device() {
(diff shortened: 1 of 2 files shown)


szabolcs-horvath:
Validated with the following images:
- FreeIPA: https://github.infra.cloudera.com/cloudbreak/cb-image-catalog/commit/497d5f249ae2a21a552fa5ff0f91ccaf0f03d473
- Prewarm: https://github.infra.cloudera.com/cloudbreak/cb-image-catalog/commit/81345da88af8ab4ff7fe77753401e03954de27b9

saltstack/base/salt/luks/bin/reopen-luks-volume.sh
```diff
@@ -50,13 +57,62 @@ decrypt_passphrase_ciphertext() {
...
+    if [[ -z "$KMS_IP_ADDRESS" ]]; then
+      echo "Failed to resolve $KMS_FIPS_ENDPOINT with any DNS server... Exiting LUKS volume reopen script with failed exit code!"
+      exit 5
+    fi
+
+    # Add entry to /etc/hosts while creating a backup of the original file
+    echo "Adding temporary entry \"$KMS_IP_ADDRESS $KMS_FIPS_ENDPOINT\" to /etc/hosts ..."
+    sed -i.bak "\$a$KMS_IP_ADDRESS $KMS_FIPS_ENDPOINT" /etc/hosts
+  else
+    echo "No need to add temporary KMS entry to /etc/hosts since this is not a FreeIPA isntance."
```
lajosrodek, line 102: Nit: `isntance` -> `instance`

lajosrodek (approved):
Nice job! LGTM, only one tiny typo :smile:

kkubina (approved):
LGTM

PR #1044, opened 2024-09-12
CB-27160 Updating pip in case of centos7 and python3.6

kkubina:


sgabor001 (approved):

TheTinkerDad:
Based on the comments on the corresponding ticket, I assume this can be closed without merging, right?

PR #1045, opened 2024-09-12
CB-27079 - Cluster side: Place nginx configs into the secret storage - Public cert

gregito:


saltstack/base/salt/luks/bin/populate-luks-volume.sh
```diff
@@ -112,7 +113,10 @@ process_gateway_locations() {
     process_location "/etc/pki/tls/certs"
     process_location_with_create "/etc/salt/pki/master" root root 700
     process_location_with_create "/srv/pillar" root root 644
-    process_nginx_locations
+    if [[ "$IS_FREEIPA" != "true" ]]
+    then
+      process_nginx_locations
+    fi
```
lajosrodek, line 119: This is wrong, the `/etc/certs` is also applicable to FreeIPA. Perhaps it makes more sense to put the `if [[ "$IS_FREEIPA" != "true" ]]` check around the `process_location_with_create "/etc/certs-user-facing"` invocation above.

lajosrodek (changes requested):

lajosrodek (approved):
Thanks, LGTM

TheTinkerDad (approved):

PR #1071, opened 2024-11-04
CB-27690: Increased image size for x86-64 images

TheTinkerDad:
Test build: https://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/5817/
Validation: not needed, the lack of free space blocked image burning and now with this branch it's now unblocked.

### DIFF
--- a/scripts/get-image-size.sh
+++ b/scripts/get-image-size.sh
@@ -5,6 +5,7 @@ source scripts/utils.sh
 CLOUD_PROVIDER=$1
 OS=$2
 STACK_VERSION=$3
+ARCHITECTURE=$4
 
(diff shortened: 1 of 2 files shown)


scripts/get-image-size.sh
```diff
@@ -5,6 +5,7 @@ source scripts/utils.sh
 CLOUD_PROVIDER=$1
 OS=$2
 STACK_VERSION=$3
+ARCHITECTURE=$3
```
sgabor001, line 8: I think there should be $4

TheTinkerDad: Yes, corrected this, thanks!

PR #1072, opened 2024-11-04
CB-27689: Fix failing Ansible installation in case of GCP (RHEL8 CIS)

TheTinkerDad:
Test build: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/5831/

### DIFF
--- a/saltstack/final/salt/cis-controls/scripts/cis_control.sh
+++ b/saltstack/final/salt/cis-controls/scripts/cis_control.sh
@@ -43,7 +43,7 @@ if [ "${CLOUD_PROVIDER}" == "Azure" ]; then
 fi
 
 #Install and download what we need for the hardening
-python3 -m virtualenv $ANSIBLE_PATH
+python3 -m virtualenv --python="/usr/bin/python3.11" $ANSIBLE_PATH
 source $ANSIBLE_PATH/bin/activate
 python3 -m pip install ansible
 yum install -y git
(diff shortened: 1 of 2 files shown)


sgabor001 (approved):

saltstack/final/salt/cis-controls/scripts/cis_control.sh
```diff
@@ -43,7 +43,7 @@ if [ "${CLOUD_PROVIDER}" == "Azure" ]; then
 fi
 
 #Install and download what we need for the hardening
-python3 -m virtualenv $ANSIBLE_PATH
+python3 -m virtualenv --python="/usr/bin/python3.11" $ANSIBLE_PATH
```
Bajzathd, line 46: I think we have a python3.11 wrapper so it could be simplified
`python3.11 -m virtualenv $ANSIBLE_PATH`

TheTinkerDad: I hope you don't mind, but I won't change it at this point. After some trial and error it turned out that neither Python 3.11 and 3.9 works with the CIS Playbook so wasted like 4-5 hours already on finding out that I'll have to use Python 3.8. This is reflected in the latest force-push and I'd rather not make another round with a minimal gain especially because this thing blocks GCP image burning.

Bajzathd (approved):

PR #1097, opened 2024-12-09
CB-28061: Reenable HTTPS for saltboot and ensure cert files are present

szabolcs-horvath:
Previously salt-bootstrap wouldn't work with HTTPS enabled, on non-gateway instances, since we do not generate the cert files which salt-bootstrap was relying on, on these instances.

Added logic for always generating the necessary certs (given that saltboot HTTPS is enabled):
- On gateway instances, we generate them using the same ca cert used to generate the certs for nginx and jumpgate
- On all other instances, we generate a ca cert and use that to generate the necessary certs

Note: The host values in the generated certs still don't matter, as there is still neither client- nor host-side verification. We only use TLS to have the channel be encrypted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,11 +161,15 @@ else
 	PYZMQ_VERSION ?= 19.0
 endif
 SALTBOOT_VERSION ?= "0.14.0"
-SALTBOOT_MINOR_VERSION = $(shell echo $(SALTBOOT_VERSION) | cut -d. -f2)
-ifeq ($(shell [[ $(SALTBOOT_MINOR_VERSION) -ge 14 ]] && echo true),true)
-	SALTBOOT_HTTPS_ENABLED ?= false
-else
-	SALTBOOT_HTTPS_ENABLED ?= false
+ifneq ($(CLOUD_PROVIDER),YARN)
+	ifneq ($(OS),centos7)
+		SALTBOOT_MINOR_VERSION = $(shell echo $(SALTBOOT_VERSION) | cut -d. -f2)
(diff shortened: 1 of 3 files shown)


szabolcs-horvath:
@TheTinkerDad 
I disabled this for CentOS7 and YCloud.

I also ran the validator for a wide range of configurations:
- [AWS - FreeIPA - RedHat8 - x86_64](http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/2181/)
- [AWS - 7.2.18 - RedHat8 - x86_64](http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/2183/)
- [AWS - 7.3.1 - RedHat8 - arm64](http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/2185/)
- [Azure - FreeIPA - RedHat8 - x86_64](http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/2199/)
- [AWS_GOV - FreeIPA - RedHat8 - x86_64](http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/2198/) (a `tearDown` fails because the secret encryption entitlement was revoked until [CB-28133](https://jira.cloudera.com/browse/CB-28133) is fixed)

I didn't run extra validators for the two configurations you mentioned above as I tested those manually.

Let me know if there are any other configurations that should be checked. I didn't want to run every single combination as it seemed excessive.

sgabor001 (approved):

TheTinkerDad (approved):

PR #1109, opened 2025-01-23
CB-27661: SSH related configs for CIS L1

sgabor001:


saltstack/final/salt/cis-controls/common.sls
```diff
@@ -145,6 +145,13 @@ sshd_harden_sshIdealTime_ClientAliveInterval:
     - repl: "ClientAliveInterval 180"
     - append_if_not_found: True
 
+sshd_harden_sshIdealTime_ClientAliveCountMax:
+  file.replace:
+    - name: /etc/ssh/sshd_config
+    - pattern: "^ClientAliveCountMax.*"
+    - repl: "ClientAliveCountMax 3"
```
TheTinkerDad, line 152: Not sure if this will work, honestly - see related tickets CB-22932 and CB-18227. I'd prefer to have the full extended validation test suite executed against this change built from a branch before approving this change. 

Also, Azure has (or at least, used to have like 1-2 years ago) its own accepted range for this config parameter - please check the value against their documentation.

sgabor001: The mentioned tickets are related to the ClientAliveInterval parameter and the e2e test issue has been solved since then. The relevant change has been deleted from the CB side since then. 
This ClientAliveInterval value is set to 180 during image burning phase as it is required by azure (also)

The **ClientAliveCountMax** is also related to the previous one. This determines the maximum number of attempts. The value 3 is a default.
Timeout value = ClientAliveInterval * ClientAliveCountMax

TheTinkerDad (approved):

PR #1130, opened 2025-03-06
CB-28730: Bump cdp-request-signer version to be able to configure TLS ciphers

szabolcs-horvath:
Previous version: `1.3.6-b2`
New version: `1.3.7-b2`

The TLS cipher configuration wasn't properly taken into consideration, as it wasn't applied to the actual cdp-request-signer server, only the client passed to the proxy request handler.


szabolcs-horvath:
Manually tested with image http://build.eng.hortonworks.com/job/cloudbreak-image-burn-salt-v3-public/7/ on mow-dev, where monitoring is enabled.

TheTinkerDad (approved):

PR #1137, opened 2025-04-16
CB-29121 Nginx host (private) certs directory & files labelling and creation of httpd_cert

romedeiro:
Files under /etc/certrs need to use the type httpd_cert


kkubina (approved):
LGTM

PR #1144, opened 2025-05-05
CB-29374: Base images are not copied to necessary regions

jszeles:
We need to distinguish a different copy phase for internal images (besides the original test phase), so the 'internal-test' phase is being introduced where a reduced region set will be used for the copy step. Basically this change just affects the azure image burning because the aws region set is the same as normal and internal cases.


kkubina (approved):
LGTM

PR #1150, opened 2025-05-28
CB-28624 CB-29106 CB-29108 Create SELinux policy for `salt-bootstrap`

lajosrodek:
* In general, to be consistent with the `cdp-xyz` convention of SELinux policy module names, all custom types have the `cdp_` prefix.
* Add SELinux policy module `cdp-common`:
  * Certain types specific to CDP services need to be referenced in multiple places. This inherently introduces an ordering dependency between two or more modules.
  * To make the said types available to all CDP-specific modules, while also keeping the implementation simple, `cdp-common` gets introduced. The only purpose of this module is to host the definitions of any types (along with their type attributes) that would be otherwise placed in some other CDP-specific module.
  * Only the `.te` file is provided, and the types provided cover both `cdp-salt` and `cdp-salt-bootstrap`.
* Add SELinux policy module `cdp-salt`:
  * Skeletal implementation for Salt services.
  * The following types are introduced:
    * `cdp_salt_cert_t`: Certificate files (and their private keys) in `/etc/salt/pki/`.
    * `cdp_salt_etc_t`: Config files in `/etc/salt/`.
    * `cdp_salt_srv_t`: All the files deployed into `/srv/` (e.g., pillars, states and some supporting files).
  * Besides a barebones `.te`, both a `.fc` and `.restorecon` file are provided.
* Add SELinux policy module `cdp-salt-bootstrap`:
  * Implicitly depends on `cdp-common` and `cdp-salt`.
  * The following types are introduced:
    * `cdp_salt_bootstrap_t`: Daemon process domain.
    * `cdp_salt_bootstrap_cert_t`: Certificate files (and their private keys) in `/etc/salt-bootstrap/certs/`.
    * `cdp_salt_bootstrap_etc_t`: Config files in `/etc/salt-bootstrap/`.
    * `cdp_salt_bootstrap_exec_t`: Executable file type for `/usr/sbin/salt-bootstrap`.
    * `cdp_salt_bootstrap_log_t`: Log file `/var/log/saltboot.log`.
    * `cdp_salt_bootstrap_port_t`: Port context type for TCP 7070 (plain HTTP) and TCP 7071 (HTTPS).
    * `cdp_salt_bootstrap_systemctl_t`: Dedicated domain for `systemctl` child processes.
    * `cdp_salt_bootstrap_unconfined_t`: Dedicated unconfined domain for certain child processes.
  * The first ~40 lines of the `.te` file was adapted from the `sepolicy generate` output (with some tweaks to properly restrict dir & file accesses). The rest is mostly handcrafted, with some help from `audit2allow`.
  * Whenever possible, child processes are launched in appropriate confined domains (via domain transitions). In particular, this is currently applicable to the following ones: `chage`, `hostname`, `sss_cache` (transitively launched by `chage` and `useradd`), `systemctl`, `useradd`.
  * Some executables are not labeled and confined by default, hence the associated child processes will be launched into the dedicated `cdp_salt_bootstrap_unconfined_t` domain (which is a variant of the built-in `unconfined_t`). This affects the following tools: `cp`, `grep`, `mv`, `ps`.
  * Close attention is paid to enable the correct labeling of all dirs & files created by `cdp_salt_bootstrap_t` or `cdp_salt_bootstrap_unconfined_t` (via appropriate file type transitions).
  * File access permissions are minimized whenever feasible. Notable exceptions are `/etc/sysconfig/network` that is not labeled properly (inhering `etc_t` of its parent `/etc/`), as well as the temporary files deployed into `/tmp/`.
  * Besides the `.te` file, `.fc`, `.portcon` and `.restorecon` ones are also provided.
* `install-cdp-policies.sh`:
  * Narrow the scope of some (environment) variables that are meant to be `local`.
  * Add handling of `.portcon` files (that are also optional, quite like `.restorecon` ones). Since `portcon` statements are not permitted in policy modules, port labeling is performed via `semanage port`.
  * Module `cdp-common` is now handled specially: It is assumed to always exist, and compiled & installed as the very first module among CDP-specific ones.
* `user-data-helper.sh`:
  * As noted in [CB-29572](https://cloudera.atlassian.net/browse/CB-29572), this script is currently run unconfined, hence most of the files it produces have incorrect file contexts.
  * As an interim solution, execute `restorecon` for `/etc/salt-bootstrap/` and `/etc/salt-bootstrap/certs/` explicitly. This workaround can later be removed once the above ticket gets fixed.
* `cis-controls/playbooks/selinux.yml`: Remove TCP 7070 (`salt-bootstrap` plain HTTP) from `http_port_t`. This port is now labeled as `cdp_salt_bootstrap_port_t`.

### DIFF
--- a/saltstack/base/salt/selinux/etc/selinux/cdp/salt-bootstrap/cdp-salt-bootstrap.te
+++ b/saltstack/base/salt/selinux/etc/selinux/cdp/salt-bootstrap/cdp-salt-bootstrap.te
@@ -1,0 +1,167 @@
+policy_module(cdp-salt-bootstrap, 1.0.0)
+
+########################################
+#
+# Declarations, domain setup
+#
+
+require {
+    type cdp_salt_bootstrap_t, cdp_salt_bootstrap_exec_t;
+}
+init_daemon_domain(cdp_salt_bootstrap_t, cdp_salt_bootstrap_exec_t)
+
+########################################
+#
+# cdp-salt-bootstrap local policy
+#
+allow cdp_salt_bootstrap_t self:fifo_file rw_fifo_file_perms;
+allow cdp_salt_bootstrap_t self:unix_stream_socket create_stream_socket_perms;
+
+require {
+    type cdp_salt_bootstrap_etc_t, cdp_salt_bootstrap_cert_t;
+}
+search_dirs_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_etc_t, cdp_salt_bootstrap_etc_t)
+search_dirs_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_cert_t, cdp_salt_bootstrap_cert_t)
+read_files_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_etc_t, cdp_salt_bootstrap_etc_t)
+read_files_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_cert_t, cdp_salt_bootstrap_cert_t)
+
+require {
+    type cdp_salt_bootstrap_log_t, var_log_t;
+}
+manage_dirs_pattern(cdp_salt_bootstrap_t, var_log_t, cdp_salt_bootstrap_log_t)
+manage_files_pattern(cdp_salt_bootstrap_t, var_log_t, cdp_salt_bootstrap_log_t)
+logging_log_filetrans(cdp_salt_bootstrap_t, cdp_salt_bootstrap_log_t, file)
+
+domain_use_interactive_fds(cdp_salt_bootstrap_t)
+
+files_read_etc_files(cdp_salt_bootstrap_t)
+
+miscfiles_read_localization(cdp_salt_bootstrap_t)
+
+########################################
+#
+# cdp_salt_bootstrap_port_t, networking
+#
+require {
+    type cdp_salt_bootstrap_port_t;
+	class tcp_socket { accept bind connect create getattr getopt listen name_bind name_connect setopt };
+}
+allow cdp_salt_bootstrap_t cdp_salt_bootstrap_port_t:tcp_socket { name_bind name_connect };
+allow cdp_salt_bootstrap_t self:tcp_socket { accept bind connect create getattr getopt listen setopt };
+corenet_tcp_bind_generic_node(cdp_salt_bootstrap_t)
+dev_read_sysfs(cdp_salt_bootstrap_t)
+kernel_read_net_sysctls(cdp_salt_bootstrap_t)
+kernel_search_network_sysctl(cdp_salt_bootstrap_t)
+
+########################################
+#
+# Child processes: chage
+#
+usermanage_domtrans_passwd(cdp_salt_bootstrap_t)
+
+########################################
+#
+# Child processes: useradd
+#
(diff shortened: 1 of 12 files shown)


saltstack/final/salt/cis-controls/playbooks/selinux.yml
```diff
@@ -13,9 +13,8 @@
         - selinux
         - CB-27855
 
-    - name: Allow httpd_t (e.g nginx) to connect to port 7070
+    - name: Allow httpd_t (e.g nginx) to connect to downstream service ports
       shell: | 
-        semanage port -a -t http_port_t -p tcp 7070
         setsebool -P httpd_can_network_connect on
```
szabolcs-horvath, line 18: (Not related to this PR)
All the selinux related commands here could probably be moved to salt states. As far as I can see they are only here because there is an OS version check for executing this playbook, but that check could also be done in a salt state.
@kkubina might be able to tell if there is any reason not to move them to the salt states.

saltstack/base/salt/selinux/etc/selinux/cdp/salt-bootstrap/cdp-salt-bootstrap.te
```diff
@@ -0,0 +1,167 @@
...
+}
+search_dirs_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_etc_t, cdp_salt_bootstrap_etc_t)
+search_dirs_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_cert_t, cdp_salt_bootstrap_cert_t)
+read_files_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_etc_t, cdp_salt_bootstrap_etc_t)
+read_files_pattern(cdp_salt_bootstrap_t, cdp_salt_bootstrap_cert_t, cdp_salt_bootstrap_cert_t)
+
+require {
+    type cdp_salt_bootstrap_log_t, var_log_t;
+}
+manage_dirs_pattern(cdp_salt_bootstrap_t, var_log_t, cdp_salt_bootstrap_log_t)
```
szabolcs-horvath, line 31: I assume this line is not strictly necessary, since we don't have a separate directory for saltboot logs inside `/var/log/`, but I am fine with it staying here.

lajosrodek: Correct. This was originally included in the `sepolicy generate` output, and I did not want to remove it. :smile:

saltstack/final/salt/cis-controls/playbooks/selinux.yml
szabolcs-horvath: Technically we could also install our policy modules through Ansible to have the SELinux related configuration in one place. I'll probably ask Kristof if he is okay with me consolidating these SELinux related configurations in one place, as they are a bit scattered right now. 😄

TheTinkerDad:
@kkubina as I explained it today, validator links are alive for 2 days best case because of the super low capacity of the Jenkins host running them, so we'll need new runs.

I've checked the catalogs for these images and I've only found the FreeIPA one for which I kicked off a new validator run:
http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3313/

TheTinkerDad (approved):

sgabor001 (approved):

PR #1168, opened 2025-07-25
CB-29490: Run salt services as confined SELinux services

szabolcs-horvath:
### The following tickets are resolved in this commit:
- CB-29490
- CB-29842
- CB-29129
- CB-29128

### The SELinux policies in this commit define the following custom SELinux domains:
- `cdp_salt_t`
- `cdp_salt_systemctl_t`
- `cdp_ipa_management_t`
- `cdp_policy_installer_t`

`cdp_salt_t`: Used by the most of processes started by the three salt services (salt-api, salt-master, salt-minion). Has fairly wide permissions, that cover our salt use cases.

`cdp_salt_systemctl_t`: The `cdp_salt_t` processes transition to this more restrictive domain, when executing `systemctl` commands.

`cdp_ipa_management_t`: Used for executing the ipa related binaries. It is more restrictive, than the salt the `cdp_salt_t` domain, and the salt processes transition to it when executing ipa commands.

`cdp_policy_installer_t`: Used for running the `install-cdp-policies.sh` script. By having a separate policy for this, we can isolate some of the SELinux related permissions to only needed for this script, so we don't have to grant it to the `cdp_salt_t` domain.

### Implementation details:
- The salt services are Python processes, so I had to create a wrapper for the Python used for salt, to avoid transitioning all Python processes to the `cdp_salt_t` domain. By using a wrapper, we can make the wrapper an entrypoint to the salt domain, without affecting any other service using the same Python executable.
- Fixed the handling of interfaces defined in the `cdp-common` module. Previously, the interfaces defined there were not available to be used in the other policies, because the Makefile used to compile the policies only took into account the interfaces present under the subdirectories of the `/usr/share/selinux/devel/include/` directory. To fix this, we copy the interface file of the `cdp-common` module into the `/usr/share/selinux/devel/include/cdp/` directory. This way, the interfaces defined in it are available to be used in all our other custom policies.
- Refactored the `install-cdp-policies.sh` script, by moving the common functions into a separate file called `policy-install-utils.sh`, which can be sourced in elsewhere, like during the installation of the `cdp-policy-installer` module.
- Refactored the previous unstructured policy used to handle nginx certs, and moved it into the `cdp-httpd` module. Also gave the custom `httpd_cert_t` a `cdp_` prefix.
- Created the `cdp-init` and `cdp-kerberos` modules, which grant a couple of permissions to existing domains, because these permissions became needed as the files got their proper labeling.
- Changed the labeling in the `cdp-postgresql` module, so that the entire `/dbfs` directory is labeled.
- Enabled the following SELinux booleans:
	- `authlogin_nsswitch_use_ldap`
	- `daemons_dump_core`
	- `domain_can_mmap_files`
	- `domain_can_write_kmsg`
	- `httpd_can_network_connect`
	- `polyinstantiation_enabled`
- Moved the SELinux salt states to the end of the `top.sls` files, so that the files created by previous salt states are present and ready to be properly labeled by the time the SELinux states run
- Removed earlier workarounds for applying file contexts, port contexts and granting permissions; and moved them to our custom policies.

### Incremental Enforcement:
**All new domains created in this commit are made permissive for now.**  This means, that whenever one of these domains would be denied an action, even in Enforcing mode, the domain will still be allowed to perform the action, and the fact that there would have been a deny will still be logged in the audit log.

This effectively allows us to confine our services on our images, without immediately breaking something if haven't considered every single possible permission that would have to be granted. This way, we can release these policies and validate them with real use cases. Of course, in an ideal world, we know every single permission that our domain would need, and we could grant them before ever releasing the policy. In the real world, like in the case of salt processes, the range of permissions we would have to grant our domain is extremely wide and also ever changing, so it's impossible to create a policy that is both restrictive and also considers every possible use case.

Releasing policies, which might not have every single necessary permission already in them, necessitates the existence of some way of validating these policies in the wild. That is what the ticket CB-29352 aims to provide. The main idea is to include validation in our image validation and E2E testing, that checks the audit log for denies, and fails the test if there are any. 

A possible problem with the above approach of checking the audit log for denies is that there are AVC checks that have so called dontaudit rules. Denies matching these dontaudit rules don't show up in the audit logs, so they could go undetected. These dontaudit rules are usually only specified for patterns that will cause denies but the denies don't actually cause any issue. However, there are also cases where the dontaudit rules hide actually function impairing denies. There are two ways of discovering the non-audited, but function impairing denies: one is the disable all dontaudit rules, and the second is to run tests in Enforcing mode (making sure the domain itself is also enforcing). The former is not ideal, as it pollutes the audit log with denies we don't actually want to permit. So as far as testing the policies, we most likely also need tests running in Enforcing mode.

## How Has This Been Tested?

- [x] Runtime image burning (per provider)
  - AWS: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8119/
  - AWS_GOV: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8138/
  - Azure: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8125/
  - GCP: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8126/
- [x] Runtime image validation (per provider)
  - AWS: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3731/
  - AWS_GOV: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3739/ 🟠
  - Azure: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3732/
  - GCP: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3733/
- [x] FreeIPA image burning (per provider)
  - AWS: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8118/
  - AWS_GOV: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8121/
  - Azure: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8123/
  - GCP: http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8124/
- [x] FreeIPA image validation (per provider)
  - AWS: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3727/
  - AWS_GOV: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3734/
  - Azure: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3728/
  - GCP: http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/3730/
- [ ] Base image burning: 
- [ ] Base image validation:


saltstack/base/salt/prerequisites/usr/bin/user-data-helper.sh
```diff
@@ -3,6 +3,8 @@
 set -ex
 set
 
+echo "The security context of this script's process is '$(id -Z)'"
```
imre-sumegi, line 6: I think I have left this line in my previous PR. :)
Anyways, I would remove this line because it is not needed anymore.

saltstack/base/salt/selinux/etc/selinux/cdp/install-cdp-policies.sh
```diff
@@ -3,83 +3,25 @@
...
-    log "Applied port contexts for CDP SELinux policy '$policy_name'"
-  else
-    log "No .portcon file found for CDP SELinux policy '$policy_name'. Skipping port context application."
-  fi
-}
+source /etc/selinux/cdp/policy-install-utils.sh
 
 main() {
-  CDP_POLICY_COMMON=common
+  log "$LOG_FILE" "The security context of this script's process is '$(id -Z)'"
```
imre-sumegi, line 11: I would remove this line because it is not needed anymore.

sgabor001 (approved):

PR #1211, opened 2025-09-12
CB-30533 - This change is to prevent waagent from mounting ephemeral volumes in Azure.

vivprabu:
This change is to prevent waagent from mounting ephemeral volumes in Azure.

1.  ResourceDisk.Format is the setting name that is related to formatting the ephemeral disk. Setting this to n will prevent Azure Linux agent from automatically formatting ephemeral disks.
2.  ResourceDisk.EnableSwap is the setting name that is related to creating and configuring a swap file on the ephemeral disk. Setting this to n will disable swap file on ephemeral disks.
3.  ResourceDisk.MountPoint is the setting name that is used to specify the mount point for the ephemeral disks. Setting this to empty will inform Azure Linux agent to not mount ephemeral disks.

## Description

Please include a summary of the change and specify which issue it addresses.

## How Has This Been Tested?

- [ ] Base image burning
- [ ] Base image validation

> **Note:**  

### DIFF
--- a/saltstack/final/salt/waagent/init.sls
+++ b/saltstack/final/salt/waagent/init.sls
@@ -19,6 +19,30 @@ set_waagent_execute_user_data:
     - repl: "Provisioning.ExecuteCustomData=y"
     - onlyif: ls /etc/waagent.conf
 
+set_waagent_resourcedisk_format:
+  file.replace:
+    - name: /etc/waagent.conf
+    - pattern: "ResourceDisk.Format.*"
+    - repl: "ResourceDisk.Format=n"
+    - append_if_not_found: True
+    - onlyif: ls /etc/waagent.conf
+
+set_waagent_resourcedisk_swap:
+  file.replace:
+    - name: /etc/waagent.conf
+    - pattern: "ResourceDisk.EnableSwap.*"
+    - repl: "ResourceDisk.EnableSwap=n"
+    - append_if_not_found: True
+    - onlyif: ls /etc/waagent.conf
+
+set_waagent_resourcedisk_mount:
+  file.replace:
+    - name: /etc/waagent.conf
+    - pattern: "ResourceDisk.MountPoint.*"
+    - repl: "ResourceDisk.MountPoint="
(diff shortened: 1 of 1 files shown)


lacikaaa (approved):
I would add `append_if_not_found: True` to the states to ensure they are always present, otherwise lgtm

vivprabu:
> I would add `append_if_not_found: True` to the states to ensure they are always present, otherwise lgtm

added the append statement Misi

vivprabu:
> One question: have you tried this, and the regex not being escaped and `*` instead of `.*` worked?
> 
> I mean this would be the ideal representation:
> 
> ```
> ^ResourceDisk\.Format=.* 
> 
> This is valid for all `repl` patterns
> ```

@pdarvasi this is the same way the other patterns are also provided in the same file. I have tried using similar patterns in other salt states.

vivprabu:
@pdarvasi thank you for pointing it out. will replace it

pdarvasi (approved):
LGTM now:)

TheTinkerDad:
It seems like image burning fails with:
``` azure-arm.arm-redhat8: local:
    azure-arm.arm-redhat8:     Data failed to compile:
    azure-arm.arm-redhat8: ----------
    azure-arm.arm-redhat8:     Rendering SLS 'final:waagent' failed: found unknown escape character
```
I guess it's the "\.", since I see no other escape sequence. I've checked some other files and seems like you don't need to escape the dot mid-string, e.g. Salt happily accepts

`pattern: "^fips.keystore.type=PKCS11"`

Please fix it on the branch and hit Retry on the two jobs I've linked above.

vivprabu:
@TheTinkerDad the builds for both AWS and Azure completed successfully. 
Azure - http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8674/ 
AWS - http://build.eng.hortonworks.com/job/cloudbreak-multi-packer-image-builder-salt-v3/8676/

TheTinkerDad (approved):

PR #1324, opened 2026-05-14
CB-30650: Add cloudera usage type lable to gcp images

jszeles:
## Description

Label cloudera-usage-type was added to the GCP images with new packer structure (instead of 'tags') because of the parameters in the tag structure are not passed to the image.

## How Has This Been Tested?

- [x] Runtime image burning (GCP) https://build.eng.cloudera.com/job/cloudbreak-multi-packer-image-builder-salt-v3/1912

> **Note:**  

### DIFF
--- a/packer.json
+++ b/packer.json
@@ -567,15 +567,21 @@
       "disk_size": "{{user `image_size`}}",
       "state_timeout": "15m",
       "tags": [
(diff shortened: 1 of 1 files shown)


kkubina (approved):
LGTM

PR #1330, opened 2026-05-28
CB-33131: Enhance ami copy job with tag copying

jszeles:
## Description

During the copy process we need to move the owner and cloudera-usage-type tags to further AMIes as well.

## How Has This Been Tested?

https://build.eng.cloudera.com/job/cloudbreak-copy-image/760/console


kkubina (approved):
LGTM

PR #1331, opened 2026-05-29
CB-28239 ipa-healtcheck should be available on new rhel9 images

daszabo:
## Description

ipa-healthcheck is a RedHat tool to check and diagnose FreeIPA. This commit introduces the ipa-healthcheck packages on new rhel9 images. The packages is should only be installed on freeipa image types.

## How Has This Been Tested?

NA, package is only installed on FreeIPA images
NA, package is only installed on FreeIPA images
- [x] FreeIPA image burning (per provider)
https://build.eng.cloudera.com/view/all/job/cloudbreak-multi-packer-image-builder-salt-v3/2000/
- [x] FreeIPA image validation (per provider)
http://ci-cloudbreak.eng.hortonworks.com/job/api-e2e-imagevalidator-v2/5799/
- [ ] Base image burning
NA, package is only installed on FreeIPA images
- [ ] Base image validation
NA, package is only installed on FreeIPA images

### DIFF
--- a/saltstack/freeipa/salt/freeipa/init.sls
+++ b/saltstack/freeipa/salt/freeipa/init.sls
@@ -48,6 +48,7 @@ freeipa-install:
         - ipa-server-trust-ad
         - bind-dyndb-ldap
         - samba-client
+        - ipa-healthcheck
 {% endif %}
 
 {% if freeipa_plugin_rpm_url %}


saltstack/base/salt/prerequisites/packages.sls
```diff
@@ -77,6 +77,9 @@ packages_install:
       - setools-console
       {% endif %}
     {% endif %}
+    {% if pillar['OS'] == 'redhat9' and pillar['CUSTOM_IMAGE_TYPE'] == 'freeipa' %}
```
lacikaaa: please move this to:
https://github.com/hortonworks/cloudbreak-images/blob/master/saltstack/freeipa/salt/freeipa/init.sls#L32

lacikaaa (changes requested):

saltstack/base/salt/prerequisites/packages.sls
daszabo: Moved it, let me retest the change

daszabo: https://build.eng.cloudera.com/job/cloudbreak-image-burn-salt-v3/1662/

daszabo: Verified the image built with these changes: `a1e5c33a-65c0-4b26-9ab5-10567d0e8d0b`

lacikaaa: why didn't you added it to the pkg section at line 44? it already has the `if` for rhel9. I don't think a separate state is required for this package

daszabo: At first I wanted to add it there, but then I was thinking that this package is not neccessarily part of, or needed for FreeIPA and the state is called `freeipa-install`. Also there are other packages with their own state. 

I don't have a strong opinion on this though, so I moved it to the suggested location.

lacikaaa (approved):

TheTinkerDad:
The title says RHEL 9, but the change itself affects both RHEL 8 and RHEL 9 images. We need to support RHEL 8 for a couple of months more, AFAIK, so this needs to be considered too. Will this work with RHEL 8 as well?

lacikaaa:
> The title says RHEL 9, but the change itself affects both RHEL 8 and RHEL 9 images. We need to support RHEL 8 for a couple of months more, AFAIK, so this needs to be considered too. Will this work with RHEL 8 as well?

It's added only for RHEL9, please check the change.
I think we're fine having this only for RHEL9, as the backend needs to be updated to use this tooling, and maybe the healthagent too, so this is just the first step.

TheTinkerDad:
> It's added only for RHEL9, please check the change. I think we're fine having this only for RHEL9, as the backend needs to be updated to use this tooling, and maybe the healthagent too, so this is just the first step.

Ok, thanks for the explanation! I've also missed that the changed line is in fact inside a bigger if block, so I don't see any reason anymore for holding this back.

TheTinkerDad (approved):

PR #1333, opened 2026-06-05
CB-32597 Update prometheus to 3.5.0

manvinderr21:
## Description

As per OBS-10200 Prometheus has vulnerabilities which need to resolve in the next FreeIPA image release, we need to update the Prometheus version because of those CVEs.

## How Has This Been Tested?

- [ ] Base image burning
- [ ] Base image validation

> **Note:**  


kkubina:
@manvinderr21 did you burn test images? please do, also rebase to latest master

PR #1352, opened 2026-07-28
Install jumpgate agent version 3.16 on both x86 and arm images

kkubina:
## Description

Install jumpgate agent ARM 3.16 on both x86 and arm images

## How Has This Been Tested?

- ~~[ ] Runtime image burning (per provider)~~
- ~~[ ] Runtime image validation (per provider)~~
  - x86 https://build.eng.cloudera.com/job/cloudbreak-multi-packer-image-builder-salt-v3/2640/
  - arm64 https://build.eng.cloudera.com/job/cloudbreak-multi-packer-image-builder-salt-v3/2641/
- ~~[ ] Base image burning~~
- ~~[ ] Base image validation~~

### DIFF
--- a/scripts/packer.sh
+++ b/scripts/packer.sh
@@ -57,9 +57,9 @@ packer_in_container() {
   # Jumpgate Agent
   if ! [[ $JUMPGATE_AGENT_RPM_URL =~ ^http.*rpm$ ]]; then
     if [[ "$ARCHITECTURE" == "arm64" ]]; then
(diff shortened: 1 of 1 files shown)


TheTinkerDad:
I suggest closing this - the normal process is that the jumpgate agent team uses the jumpgate-agent-dev-release job for creating their own test images, then copying those into the necessary regions, then triggering their own e2e tests via quasar/quanta. If their testing succeeds, they'll open a PR for this - see https://github.com/hortonworks/cloudbreak-images/pull/1289

PR #1362, opened 2026-09-01
CB-34272 Increase OpenStack instance-ID retry timeout and add random-UUID fallback

sodre90:
## Description

CB-34272: On OpenStack (Platform9), the instance-metadata read during boot
(`curl http://169.254.169.254/openstack/latest/meta_data.json`) sometimes
doesn't succeed within the previous 10 attempts x 5s (~50s) retry window,
aborting boot init entirely (script runs under `set -ex`).

Scoped to the `OPENSTACK` branch of `get_instance_id()` only — AWS/Azure/GCP
behavior is unchanged:

- Retry window extended to 60 attempts x 5s (~5 min).
- The `curl` call is now bounded with `--connect-timeout 5 --max-time 10`,
  so a hung metadata endpoint fails fast and retries instead of stalling
  the whole loop.
- After exhausting retries, falls back to a random UUID (from
  `/proc/sys/kernel/random/uuid`, or `uuidgen`) instead of returning 1 and
  aborting boot.

Note: the only downstream consumer of `INSTANCE_ID` is
`ccmv2-helper.sh`'s `BACKEND_ID`. If OpenStack ever actually falls back to
a random UUID, that node's CCMv2 backend ID won't match anything the
control plane expects — degraded CCMv2 registration is accepted here as
preferable to a bricked boot.

## How Has This Been Tested?

- [ ] Runtime image burning (per provider) — not run; this touches only the shared boot-time helper script, exercised at instance first-boot rather than image burn time.
- [ ] Runtime image validation (per provider) — same as above.
- [ ] FreeIPA image burning (per provider) — not applicable; change is provider-scoped to OpenStack.
- [ ] FreeIPA image validation (per provider) — not applicable; change is provider-scoped to OpenStack.
- [x] Base image burning — manually triggered an OpenStack redhat9 base image burn against Platform9 (iopscloud/cloudbreak project) to sanity-check the surrounding change locally; not a Jenkins-linked run.
- [ ] Base image validation — pending: still need to boot an instance from the resulting image on OpenStack to exercise the retry/fallback path itself under real metadata-service conditions.


kergyik:
https://build.eng.cloudera.com/job/cloudbreak-multi-packer-image-builder-salt-v3/2935/

kergyik (approved):
LGTM

